PR #3, opened 2022-06-13
Front start

matudelic:
Adds initial support for necessary techs (i.e. node modules)

-imports (and successfully uses) Helsinki Design System-components


vincit-matu:
...aukes vahingossa siviiliaccountilla 😊 No, käypä se kai näinkin (?)

package.json
```diff
@@ -1,6 +0,0 @@
-{
-  "dependencies": {
-    "postcss-loader": "^7.0.0",
-    "sass": "^1.52.3"
-  }
-}
```
ranta: Instead of making a separate commit for this, I would instead squash this to the original commit to keep history cleaner by keeping "noise" out :)

ranta:
Try to aim for max 72 characters on the first row of commit messages, GitHub wraps messages longer than that to a new line (by adding ellipsis a ...), which looks a bit ugly :D

ranta (approved):
No need to consider these comments blocking, just something to keep in mind in the future

package.json
vincit-matu: Kuitti

PR #10, opened 2022-07-04
Custom error handling

toikarin-vincit:
3ff113f: error handling changes
 - implement custom handlers for 400, 404, 405, 406, 415, 500
 - 400 is implemented according to OpenAPI
 - disabled localization and internationalization, at least for now
  - this way error strings are not localized
 - added OpenAPI definitions for 406 and 415
 - added an simple OpenAPI definition for creating a housing company
   - this helps with testing 400 errors
---
6e6567b: minor style fixes
---
6de444c: Makefile: run `make format` on `make [all]`

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,6 +1,6 @@
 .PHONY = all build format tests dump
 
-all: tests build
+all: format tests build
 
 build:
 	./manage.py collectstatic --clear --no-input --verbosity=0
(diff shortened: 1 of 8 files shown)


backend/config/settings.py
```diff
@@ -121,8 +121,8 @@
 LOCALE_PATHS = ["./templates/locale"]
 
 TIME_ZONE = "Europe/Helsinki"
-USE_I18N = True
-USE_L10N = True
```
ranta, line 125: Why disable these? 🤔

ranta (dismissed):

backend/config/settings.py
toikarin-vincit: I originally changed those so the returned error messages are not translated as stated in the commit message. But they are not needed anymore so I reverted the change.

ranta (approved):

PR #12, opened 2022-07-05
Add env `HITAS_TESTS_NO_DOCKER` for disabling docker use in tests

toikarin-vincit:
- local PostgreSQL is required in this case


ranta (dismissed):
❤️

ranta (approved):

PR #13, opened 2022-07-05
HousingCompany Create serializer

ranta:
- Add Serializer for creating and updating HousingCompanies
- Add factories for all models to make future testing easier
- Improve existing serializers a bit
- Few small misc changes and fixes


backend/hitas/views/helpers.py
```diff
@@ -58,10 +60,51 @@ def get_object(self):
...
+
+class UUIDRelatedField(SlugRelatedField):
+    def __init__(self, **kwargs):
+        super().__init__(slug_field="uuid", **kwargs)
+
+    def to_representation(self, obj):
+        return getattr(obj, self.slug_field).hex
+
+    def to_internal_value(self, data):
+        data = UUID(hex=data)
```
toikarin-vincit, line 90: handle the cases when request does not contain a valid UUID

backend/hitas/tests/apis/test_api_housing_company.py
```diff
@@ -0,0 +1,100 @@
...
+        "property_manager": property_manager.uuid.hex,
+        "state": "not_ready",
+        "sales_price_catalogue_confirmation_date": "2022-01-01",
+    }
+
+    response = api_client.post(reverse("hitas:housing-company-list"), data=data, format="json")
+    # validate_openapi(response)  # TODO
+
+    assert response.status_code == status.HTTP_201_CREATED
+    assert response.json()["id"] == HousingCompany.objects.first().uuid.hex
```
toikarin-vincit, line 55: this now returns same kind of object as the request was. as an example:
request:
```
...
"developer": "06224336de274d87853afb950ba5ecc8",
...
```

response:
```
...
"developer": "06224336de274d87853afb950ba5ecc8",
...
```
when we read it with `GET`:
```
    "developer": {
        "code": "007",
        "description": null,
        "value": "Helsingin Sato Oy"
    },
```

The `POST` response should return the same as `GET`. Also the ID is not available in the response

toikarin-vincit (commented):
currently I would like to see still few things:
 - rebase to current master
 - fix tests (id's as in OpenAPI)
 - add more tests, especially error cases

PR #16, opened 2022-07-07
Changes to dev pipelines

jannetasa:


toikarin-vincit (approved):
end result looks OK, commits are bit funky

PR #20, opened 2022-07-07
docker changes for running in Platta

toikarin-vincit:
821e876: Dockerfile: add pcre support
 - previously logs stated `!!! no internal routing support, rebuild with pcre support !!!`
 - install libpcre3 to get rid of this
---
3d6e295: Dockerfile: run compilemessages during the image building
 - remove it from entrypoint.sh, call it on building instead
   - no need to write files during runtime anymore
---
6a907e0: uwsgi: expose 8080 as HTTP
---
5bad659: add django-health-check
 - GET /healthz now responds to health check


ranta (approved):
LGTM 👨‍⚕️

PR #39, opened 2022-08-18
Various fixes

toikarin-vincit:
ebe84ea: backend: fix flaky test__api__apartment__filter test
 - don't generate apartment numbers from the same range as the ApartmentFactory
---
05c7ee9: backend: add missing migration
---
ca7fc9a: backend/frontend: rename address.street -> address.street_address
 - it's a better name and aligned with the filter
---


ranta (approved):
👍

PR #52, opened 2022-08-30
Select postcode

vincit-matu:
Updates postal code selection in housing company creation

- adds relevant models and queries
- also fixes some react warnings (left over/out code)

### DIFF
--- a/frontend/src/common/models.ts
+++ b/frontend/src/common/models.ts
@@ -228,7 +228,7 @@ export interface PageInfo {
 export interface IPostalCode {
     value: string;
     city: string;
-    cost_area: number;
+    cost_area: 1 | 2 | 3 | 4;
 }
 
 // List response interfaces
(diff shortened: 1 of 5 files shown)


frontend/src/common/models.ts
```diff
@@ -181,6 +181,12 @@ export interface PageInfo {
     };
 }
 
+export interface IPostalCode {
+    value: string;
+    city: string | null;
+    cost_area: string;
```
toikarin-vincit, line 187: this is an integer between 1-4

frontend/src/common/models.ts
```diff
@@ -181,6 +181,12 @@ export interface PageInfo {
     };
 }
 
+export interface IPostalCode {
+    value: string;
+    city: string | null;
```
toikarin-vincit, line 186: this cannot be `null`

vincit-matu: Updated accordingly

frontend/src/common/models.ts
vincit-matu: Updated to be of type 1 | 2 | 3 | 4.

toikarin-vincit (approved):

PR #60, opened 2022-09-09
API error message improvements (continued) and fix for code order

toikarin-vincit:
bb0a22e: backend: improve error message on share related fields
 - instead of error `field: "non_field_errors", message: "share_number_start must not be greater than share_number_end"`
   now return `field: "shares", message: "'shares.start' must not be greater than 'shares.end'."`
 - add apartment constraint that checks that either both

---

5051231: backend: fix code order
 - previously fetching codes (e.g. `GET /financing-methods` returned
   the list of codes without taking `order` into account. This has now
   been fixed and codes are returned in correct order

---

b141d58: backend: initial.json: reorder financing methods
 - but "Uusi Hitas" first in the list

---

4d294b3: backend: token management command: add support for revoking the token
 - `--key` renamed to `--set`
 - `--revoke` added

---


miqu (approved):
Ei ilmeisesti haittaa, että order on sama 0?

toikarin-vincit:
> Ei ilmeisesti haittaa, että order on sama 0?

eip

PR #62, opened 2022-09-09
backend: hotfix for migration 0017

toikarin-vincit:
- also migrate official names, not only display names


vincit-matu (approved):
Express-approved!

PR #68, opened 2022-09-22
Improvements

toikarin-vincit:
5ad688b: backend: update openapi-core

---

62a7336: backend: update dependencies

---

b87997d: frontend: update dependencies

---

3052c78: backend: serializers: use NotImplementedError instead of empty function

---

d99aae5: backend: oracle-migration: fix truncation after soft-delete changes

---

ef79627: backend: migration: fix non-unique name
 - name is now required to be unique. add a simple workaround for
   handling it. Data will be fixed so this is not needed in a long run

---

c731d4f: backend: add support for housing company/apartment improvements
 - not possible to add/remove yet, just read an empty lists. thus
   everything is marked read only for now

---

4a04725: backend: oracle-migration: separate oracle schema into multiple files

---

85344d3: backend: oracle-migration: migrate housing company/apartment improvements

---


backend/hitas/models/apartment.py
```diff
@@ -116,3 +116,30 @@ class Meta:
...
+    apartment = models.ForeignKey("Apartment", on_delete=models.CASCADE, related_name="market_price_improvements")
+    name = models.CharField(max_length=128)
+    completion_date = models.DateField()
+    value = models.PositiveIntegerField(default=0)
+
+
+class DepreciationPercentage(Enum):
+    ZERO = 0.0
+    TWO_POINT_FIVE = 2.5
+    TEN = 10.0
```
ranta, line 131: Is the depreciation percentage always one of these? 🤔

backend/hitas/models/housing_company.py
```diff
@@ -99,3 +99,21 @@ class Meta:
 
     def __str__(self):
         return self.display_name
+
+
+class HousingCompanyMarketPriceImprovement(HitasModel):
```
ranta, line 104: Since there are four `Improvement` models, i could have made sense to have an abstract parent model these inherit from

backend/hitas/views/apartment.py
```diff
@@ -137,6 +151,11 @@ def create_links(instance: Apartment) -> Dict[str, Any]:
     }
 
 
+class ApartmentImprovementSerializer(serializers.Serializer):
+    market_price_index = ImprovementSerializer(many=True, source="market_price_improvements")
+    construction_price_index = ImprovementSerializer(many=True, source="construction_price_improvements")
```
ranta, line 156: Is the `depreciation_percentage` field not needed here?

ranta (approved):

backend/hitas/views/apartment.py
toikarin-vincit: yep, it's not needed :)

backend/hitas/models/housing_company.py
toikarin-vincit: I originally did that, but the foreign key caused some pain (2 of them are joined to housing company and 2 to apartments + `related_name` causes some nasty names). Back then these models were a bit different but now I agree the three common fields could move moved to an abstract base class

toikarin-vincit: I added one for commit to address this

ranta (approved):
LGTM 🚀

PR #70, opened 2022-09-22
Housing Company Legacy ID filtering

ranta:
refs. HT-249


backend/hitas/views/housing_company.py
```diff
@@ -42,6 +42,7 @@ class HousingCompanyFilterSet(HitasFilterSet):
     postal_code = HitasPostalCodeFilter(field_name="postal_code__value")
     property_manager = HitasCharFilter(field_name="property_manager__name", lookup_expr="icontains")
     developer = HitasCharFilter(field_name="developer__value", lookup_expr="icontains")
+    legacy_id = HitasCharFilter(lookup_expr="iexact")
```
toikarin-vincit, line 45: `legacy_id` is numeric so `iexact` does not really make sense. but maybe not the currect datatype either

toikarin-vincit (commented):
`legacy_id` does not really suit the purpose anymore after follow-up discussions:
 - we need this id for newly created housing companies too (for archiving purposes)
 - datatype is currently a string: we need a short numeric value (for archiving purposes)
 - name `legacy_id` does not make sense if it will be used
 
 so backend would need at least the following:
  - rename the field
  - change the type to numeric(?)
  - generate the id for all the created housing companies
  - handle migration gracefully, next generated id should be the max id from the migration + 1

miqu:
> `legacy_id` does not really suit the purpose anymore after follow-up discussions:
> 
> * we need this id for newly created housing companies too (for archiving purposes)
> * datatype is currently a string: we need a short numeric value (for archiving purposes)
> * name `legacy_id` does not make sense if it will be used
> 
> so backend would need at least the following:
> 
> * rename the field
> * change the type to numeric(?)
> * generate the id for all the created housing companies
> * handle migration gracefully, next generated id should be the max id from the migration + 1

Yeah, the name should be changed later,
I think I put generation in a separate ticket in jira.

So if we get filtering and ui changes we can already use it and it adds value too :)

backend/hitas/views/housing_company.py
ranta: The data stored in there is numeric, but the field is a CharField, so it doesn't really matter if its `iexact` or `exact`.

toikarin-vincit:
> So if we get filtering and ui changes we can already use it and it adds value too :)

the problem is that this gives us no value currently. none of the housing companies have `legacy_id` set so there's no way to use this

backend/hitas/views/housing_company.py
toikarin-vincit: `iexact` will result of `UPPER(legacy_id) = ?` instead of `legacy_id = ?`. From current data size perspective it does not matter that much but it's slower: conversion and if the column is indexed it would need to be indexed with `UPPER`, otherwise it will not be used

backend/openapi.yaml
```diff
@@ -2376,17 +2376,17 @@ components:
...
           readOnly: true
-        legacy_id:
-          description: Old legacy ID for this housing company
-          type: string
+        archive_id:
+          description: Archival ID for this housing company
+          type: integer
           readOnly: true
-          nullable: true
+          nullable: false
```
toikarin-vincit, line 2383: it's OK to remove this instead

backend/openapi.yaml
```diff
@@ -2376,17 +2376,17 @@ components:
           format: date
           nullable: true
           readOnly: true
-        legacy_id:
-          description: Old legacy ID for this housing company
-          type: string
+        archive_id:
+          description: Archival ID for this housing company
+          type: integer
```
toikarin-vincit, line 2381: let's add `minimum: 0`

frontend/src/features/housingCompany/HousingCompanyListPage.tsx
```diff
@@ -100,6 +102,14 @@ const HousingCompanyFilters = ({filterParams, setFilterParams}): JSX.Element =>
                 filterParams={filterParams}
                 setFilterParams={setFilterParams}
             />
+            <FilterIntegerField
+                label="Yhtiön arkisto tunnus"
```
toikarin-vincit, line 106: arkistotunnus

frontend/src/features/housingCompany/HousingCompanyListPage.tsx
```diff
@@ -100,6 +102,14 @@ const HousingCompanyFilters = ({filterParams, setFilterParams}): JSX.Element =>
...
+                minLength={3}
```
toikarin-vincit, line 107: this should be 1

frontend/src/features/housingCompany/HousingCompanyDetailsPage.tsx
```diff
@@ -123,6 +123,10 @@ const HousingCompanyDetailsPage = () => {
                                             label="Talotyyppi"
                                             value={data.building_type.value}
                                         />
+                                        <DetailField
+                                            label="Yhtiön arkisto tunnus"
```
toikarin-vincit, line 127: arkistotunnus

toikarin-vincit (approved):

PR #74, opened 2022-09-28
Change Apartment filter to work with shorter input

ranta:
- This allows for passing the birthdate without the end-part and still getting results

refs. HT-118 Comments


toikarin-vincit:
I think this is not needed anymore after https://github.com/City-of-Helsinki/hitas/pull/80 :thinking:

PR #78, opened 2022-09-29
S5 minifixes

vincit-matu:
Several small fixes to views, adding buttons and restructuring the creation process as per requests from sprint 5 demo


ranta (approved):
LGTM 🚀 

In future commits, could you limit the commit message length to 72 characters on the first line, so that end of the message does not have ellipsis added to it :)

vincit-matu:
> LGTM 🚀
> 
> In future commits, could you limit the commit message length to 72 characters on the first line, so that end of the message does not have ellipsis added to it :)

Sure, I'll try to keep the rambling to the following section in the message! :D

PR #80, opened 2022-09-30
backend: rename 'person' to 'owner'

toikarin-vincit:
- owner can be a person but could also be for example a company
 - drop separate last_name and first_name. only have name
   - this helps migration (old system only has name) but also makes sense in a context of company
 - rename social_security_number to identifier
   - companies do not have social security number, they use business id instead


frontend/src/features/apartment/ApartmentListPage.tsx
```diff
@@ -33,7 +33,7 @@ const ApartmentListItem = ({
 }: ApartmentListItemProps): JSX.Element => {
     // Combine ownerships into a single formatted string
     const ownershipsString = ownerships
-        .map((o) => `${o.owner.last_name} ${o.owner.first_name} (${o.owner.social_security_number})`)
+        .map((o) => `${o.owner.name} (${o.owner.identifier})`)
         .join(", ");
```
ranta, line 37: Can you run `yarn run prettier-format` to fix the formatting here, as the variable names are now shorter, prettier wants it to be single-row and raises an issue about it :)

ranta (changes requested):
Two small changes needed, otherwise looks good

backend/hitas/migrations/0022_rename_person_to_owner.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.2.15 on 2022-09-29 13:17
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('hitas', '0021_improvement_order'),
```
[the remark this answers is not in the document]
toikarin-vincit: done

frontend/src/features/apartment/ApartmentListPage.tsx
toikarin-vincit: done

ranta (approved):
LGTM 🚀

PR #81, opened 2022-10-03
Implement listing/reading/updating indices

toikarin-vincit:
d00c355: backend: add support for handling exceptions with a single error
 - handle case when error is not a list but instead a single error
 - throw NotImplementedException if unexpected case happens

---

af4ac30: backend: add support for `max_decimal_place` errors
 - this occurs when field has more than two digits

---

9af54d5: backend: HitasAPIClient: allow disabling OpenAPI validation

---

8571f49: backend: implement listing/reading/updating indices
 - Listing of indices (`GET /indices/{index}`) implemented
   - index can be one of:
     - max-value-index
     - market-price-index
     - market-price-index-pre-2005
     - construction-price-index
     - construction-price-index-pre-2005
 - Reading of a single index (`GET /indices/{index}/{month}`) implemented
 - Updating/deleting of a single index (`/indices/{index}/{month}`) implemented
 - Removed unnecessary fields for indices
   - `deleted`/`deleted_by_cascade` are unnecessary as fields are not removed
   - `month` is a better primary key as `id`

---


backend/hitas/tests/apis/helpers.py
```diff
@@ -57,7 +57,12 @@ def path_pattern(self) -> Optional[str]:
...
 class HitasAPIClient(APIClient):
-    def generic(self, method, path, data="", content_type="json", secure=False, **kwargs) -> Response:
+    def generic(
+        self, method, path, data="", content_type="json", secure=False, openapi_validate=True, **kwargs
+    ) -> Response:
         response: Response = super().generic(method, path, data, content_type, secure=False, **kwargs)
-        validate_openapi(response)
+
+        if openapi_validate:
+            validate_openapi(response)
```
ranta, line 66: Why would validating the api not be wanted? 🤔

backend/hitas/tests/factories/indices.py
```diff
@@ -0,0 +1,49 @@
...
+faker = Faker(locale="fi_FI")
+
+
+class AbstractIndexFactory(DjangoModelFactory):
+    class Meta:
+        django_get_or_create = ("value",)
+        abstract = True
+
+    month = fuzzy.FuzzyDate(start_date=datetime.date(1990, 1, 1))
+    value = fuzzy.FuzzyDecimal(low=0)
```
ranta, line 24: Should this be `django_get_or_create = ("month",)` instead?
There may be a case where the same value for month is generated, which would cause a duplicate pk exception

backend/hitas/tests/factories/indices.py
```diff
@@ -0,0 +1,49 @@
...
+
+faker = Faker(locale="fi_FI")
+
+
+class AbstractIndexFactory(DjangoModelFactory):
+    class Meta:
+        django_get_or_create = ("value",)
+        abstract = True
+
+    month = fuzzy.FuzzyDate(start_date=datetime.date(1990, 1, 1))
```
ranta, line 23: It could be an issue that the `month` value is not always the first day of a month, since when the `month` is serialised the day is dropped off and duplicate `month` values may be returned from the endpoint 🤔

backend/hitas/tests/apis/test_api_indices.py
```diff
@@ -0,0 +1,222 @@
...
+@pytest.mark.django_db
+def test__api__indices__create(api_client: HitasAPIClient, index):
+    response = api_client.post(reverse(f"hitas:{index}-list"), data={}, openapi_validate=False)
+    assert response.status_code == status.HTTP_405_METHOD_NOT_ALLOWED, response.json()
+    assert response.json() == {
+        "error": "method_not_allowed",
+        "message": "Method not allowed",
+        "reason": "Method Not Allowed",
+        "status": 405,
+    }
```
ranta, line 106: Since creating is now allowed, how are the indices created? 🤔

backend/hitas/tests/apis/helpers.py
toikarin-vincit: this is currently used when testing an endpoint that does not exist (then openapi validation would fail as well)

backend/hitas/tests/factories/indices.py
toikarin-vincit: true. will fix

backend/hitas/tests/apis/test_api_indices.py
toikarin-vincit: it's possible to fetch all the months that have values with `GET /indices/{index}`.
It's possible to query a specific month with `GET `/indices/{index}/month`. It will return a value or `null` if not set.
It's possible to update a specific month with `PUT `/indices/{index}/month` and set `value` to any value or `null`.

The idea here is that all the months "exist". Values for them might or might not.

backend/hitas/tests/factories/indices.py
toikarin-vincit: fixed

backend/hitas/tests/factories/indices.py
toikarin-vincit: fixed by removing it, it was unnecessary

backend/hitas/tests/apis/test_api_indices.py
ranta: All righty, makes sense 😄

ranta (approved):

PR #83, opened 2022-10-03
Make HousingCompany fields non-required

ranta:
Previously we had some fields marked as required, when they shouldn't be, this PR addresses the issue.

refs. HT-30 comments, HT-31 comments


backend/openapi.yaml
```diff
@@ -2270,7 +2269,8 @@ components:
         business_id:
           description: Official business ID
           type: string
-          pattern: '^\d{7}-\d$'
+          nullable: true
+          pattern: '(^$|^\d{7}-\d$)'
```
toikarin-vincit, line 2273: This would be much cleaner if the value is either a valid business id OR null. now it can be `null`, `""` or a valid id

backend/hitas/tests/apis/test_api_housing_company.py
```diff
@@ -447,6 +447,9 @@ def test__api__housing_company__create(api_client: HitasAPIClient, minimal_data:
     if minimal_data:
         data.update(
             {
+                "business_id": "",
+                "primary_loan": "",
```
toikarin-vincit, line 451: this seems to be aligned with other fields but I feel it's strange to allow empty string `""` for a numeric field :thinking:

backend/hitas/migrations/0025_required_fields.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 3.2.15 on 2022-09-28 15:19
```
toikarin-vincit, line 1: `housing_company_required_fields` would be a bit better name for this file

ranta: Renamed :)

vincit-matu (approved):
Front checks out 👍

PR #84, opened 2022-10-06
OpenAPI request validation

toikarin-vincit:
3350e61: backend: remove warning on test class
 - fix warning `cannot collect test class 'TestClass' because it has a
   __init__ constructor`

---

7a5329a: backend: update dependencies

---

5c7374d: frontend: dependency updates

---

ef88903: backend: implement OpenAPI request validation
 - now requests are also validated along with the responses
 - only used in tests but helps us to verify the OpenAPI definitions and
   test cases are valid
 - note: this will break a lot of tests, future commit will fix
   those
 - don't validate requests when the request is on purpose invalid
   - when we expect 400 Bad Request it's OK to ignore the request
     validation

---

5215a25: backend: OpenAPI: add `additionalProperties: false`
 - annoying workaround to force openapi-core to verify the default
   behaviour
 - causes test failures, will be fixed in future commit

---

2db454d: backend: OpenAPI / test fixes based on the request validation
 - fixes tests broken after request validation implemented in ef88903
   and 5215a25

---

70a79bb: backend: validate_openapi: remove unnecessary workaround

---

bc4a0a9: backend: validate_openapi: print better error messages on validation errors

---

3c6c5d2: backend: fix OpenAPI definitions on different searches

---

### DIFF
--- a/backend/hitas/tests/apis/helpers.py
+++ b/backend/hitas/tests/apis/helpers.py
@@ -10,7 +10,6 @@ from openapi_core.unmarshalling.schemas import oas30_request_schema_unmarshaller
 from openapi_core.validation.request.protocols import Request
 from openapi_core.validation.request.validators import RequestValidator
 from openapi_core.validation.response import openapi_response_validator
-from rest_framework import status
 from rest_framework.response import Response
 from rest_framework.test import APIClient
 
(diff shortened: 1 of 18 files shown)


backend/hitas/tests/apis/helpers.py
```diff
@@ -58,7 +58,16 @@ def validate_openapi(
...
+        handle_result(result)
+
+
+def handle_result(result):
+    for error in result.errors:
+        if isinstance(error, InvalidSchemaValue):
+            for suberror in error.schema_errors:
+                print(suberror)
+
+    result.raise_for_errors()
```
ranta, line 70: ❤️

ranta (approved):
🚀

PR #85, opened 2022-10-06
S5 testing fixes

vincit-matu:
Contains fixes for the issues found in testing
Includes frontend support for number of rooms for apartments (both in creation/edit and details view - with dummy data)
Includes frontend support for comparison of realised and initial acquisition price in company details page
Resolves HT-33, HT-36, HT-109, HT-118 and HT-156 for second round of testing.

### DIFF
--- a/backend/hitas/views/owner.py
+++ b/backend/hitas/views/owner.py
@@ -2,7 +2,7 @@ from rest_framework import status
 from rest_framework.response import Response
 
 from hitas.models import Owner, Ownership
-from hitas.views.utils import HitasCharFilter, HitasFilterSet, HitasModelSerializer, HitasModelViewSet, HitasSSNFilter
+from hitas.views.utils import HitasCharFilter, HitasFilterSet, HitasModelSerializer, HitasModelViewSet
 
 
 class OwnerSerializer(HitasModelSerializer):
@@ -18,7 +18,7 @@ class OwnerSerializer(HitasModelSerializer):
 
 class OwnerFilterSet(HitasFilterSet):
     name = HitasCharFilter(lookup_expr="icontains")
-    identifier = HitasSSNFilter(lookup_expr="iexact")
+    identifier = HitasCharFilter(lookup_expr="icontains")
     email = HitasCharFilter(lookup_expr="iexact")
 
     class Meta:
--- a/frontend/src/common/components/formInputField/FormInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormInputField.tsx
@@ -35,6 +35,11 @@ type FormInputFieldProps = {
           inputType?: "text" | "textArea" | "postalCode" | "number" | "date";
       }
     | {
+          inputType: "number";
+          unit?: string;
+          fractionDigits?: number;
+      }
+    | {
           inputType: "select" | "combobox";
           options: {
               label: string;
--- a/frontend/src/common/components/formInputField/FormInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormInputField.tsx
@@ -149,11 +149,14 @@ export default function FormInputField({
         }
 
         return (
-            <FormDropdownInputField
-                {...commonProps}
-                searchable={inputType === "combobox"}
-                {...rest}
-            />
+            <div className={`input-field input-field--dropdown${required ? " input-field--required" : ""}`}>
+                <FormDropdownInputField
+                    {...commonProps}
+                    searchable={inputType === "combobox"}
+                    required={required}
+                    {...rest}
+                />
+            </div>
         );
     }
     if (inputType === "date") {
--- a/frontend/src/common/components/formInputField/FormInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormInputField.tsx
@@ -32,7 +32,7 @@ type FormInputFieldProps = {
     className?: string;
 } & (
     | {
-          inputType?: "text" | "textArea" | "postalCode" | "number" | "date";
+          inputType?: "text" | "textArea" | "postalCode" | "date";
       }
     | {
           inputType: "number";
--- a/frontend/src/common/models.ts
+++ b/frontend/src/common/models.ts
@@ -72,8 +72,8 @@ export interface IHousingCompanyWritable {
     developer: {id: string};
     property_manager: {id: string};
     acquisition_price: {
-        initial: number | null;
-        realized: number | null;
+        initial: number;
+        realized: number;
     };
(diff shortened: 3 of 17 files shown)


frontend/src/features/apartment/ApartmentCreatePage.tsx
```diff
@@ -170,18 +172,18 @@ const ApartmentCreatePage = () => {
...
                         />
-                    </div>
-                    <div className={"row"}>
                         <FormInputField
                             inputType="number"
                             fractionDigits={2}
                             label="Pinta-ala"
                             fieldPath="surface_area"
-                            required
+                            required={true}
```
ranta: These two are equivalent :)

frontend/src/common/models.ts
```diff
@@ -72,8 +72,8 @@ export interface IHousingCompanyWritable {
     developer: {id: string};
     property_manager: {id: string};
     acquisition_price: {
-        initial: number | null;
-        realized: number | null;
+        initial: number;
+        realized: number;
```
ranta, line 76: `acquisition_price.initial` change is ok.
`acquisition_price.realized` should be nullable, there may be cases where the company is not yet finished, so the realizer price is unknown.

backend/hitas/views/owner.py
```diff
@@ -2,7 +2,7 @@
 from rest_framework.response import Response
 
 from hitas.models import Owner, Ownership
-from hitas.views.utils import HitasCharFilter, HitasFilterSet, HitasModelSerializer, HitasModelViewSet, HitasSSNFilter
+from hitas.views.utils import HitasCharFilter, HitasFilterSet, HitasModelSerializer, HitasModelViewSet
```
ranta, line 5: Could be squashed with the commit, in which you made changes to the same file

frontend/src/features/apartment/ApartmentCreatePage.tsx
```diff
@@ -167,7 +167,6 @@ const ApartmentCreatePage = () => {
                             inputType="number"
                             label="Kerros"
                             fieldPath="address.floor"
-                            required
```
ranta, line 168: The `required` attributes are not unnecessary?

ranta (changes requested):

frontend/src/features/apartment/ApartmentDetailsPage.tsx
```diff
@@ -70,26 +73,56 @@ const ApartmentDetailsPage = (): JSX.Element => {
...
+                                </span>
+                            </div>
+                        </div>
+                        <label className="card-heading">Vahvistettu enimmäishinta</label>
+                        <p className="confirmed-price">250000€</p>
+                        <Button
+                            className="button-confirm"
+                            theme="black"
+                            size="small"
+                            style={{display: "none"}} // Hidden until it's
```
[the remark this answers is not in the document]
vincit-matu: ..haha, indeed! The missing word is: needed :D 
Thought it might be good to not have the button there to confuse the client, while its operation is still a bit of a mystery (..and as it does nothing). Just wanted to have a readiness for the button in question.

frontend/src/features/apartment/ApartmentDetailsPage.tsx
```diff
@@ -50,6 +50,9 @@ const SalesCondition = ({
 
 const ApartmentDetailsPage = (): JSX.Element => {
     const params = useParams();
+    const [priceMH] = useState(210000);
+    const [priceRH] = useState(250000);
+    const [priceRNH] = useState(130000);
```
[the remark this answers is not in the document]
vincit-matu: That's very true. I was building a system in the front to check which of the three values was highest in order to highlight it (with an onClick changing those values to test that it worked), and that needed persistent values over re-renders. In the end I thought it was a bit unnecessary (and my solution was quite hacky) so I left that out here.
Couldn't these just be substituted with the proper values from the API - once they exist there (like we'd have to do with dynamic vs hard-coded values anyway), as they don't cause any harm as they are?

frontend/src/common/components/formInputField/FormInputField.tsx
```diff
@@ -34,6 +34,11 @@ type FormInputFieldProps = {
     | {
           inputType?: "text" | "textArea" | "postalCode" | "number" | "date";
       }
+    | {
+          inputType: "number";
```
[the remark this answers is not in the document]
vincit-matu: Fair point!

frontend/src/common/models.ts
vincit-matu: Had a chat about the acquisition prices with Tomi prior to making these changes. These two values will change to `acquisition_price` and `summary.realized_acquisition_price` anyway, once those values are provided by the API  - as now the realized one is not the calculated sum of the prices of all apartments for the housing company in question.
I also changed the initial values to be 0 instead of null in the blank `formData` contents which the creation form defaults to, so that the comparison wouldn't cause errors (and so that the details page wouldn't display "null" as a value - as the real realized value for a company with no apartments is 0.
Should this still be allowed to be null?

backend/hitas/views/owner.py
vincit-matu: Right you are - this was more of a reaction to the tests not passing, and me not being too familiar with the backend. :)

frontend/src/features/apartment/ApartmentCreatePage.tsx
vincit-matu: Quoting Saila's comment to HT-33: "Asunnon luominen ei onnistu pelkkiin tähdellä merkittyjen täyttämisellä, myös ensisijainen laina täytyy täyttää. Pakollisiksi kentiksi riittää asunnon luontivaiheessa pelkästään Rakennus, Rappu, asunnon nro, asuntotyyppi."

frontend/src/features/apartment/ApartmentCreatePage.tsx
vincit-matu: ..also, should be fixed in a later commit. Or at least I couldn't find a `required={true}` in `ApartmentCreatePage.tsx` anymore.

frontend/src/common/models.ts
vincit-matu: ..or well, anyway I added the `null` option back to the model now, for `acquisition_price.realized`, just in case it would present some problems which I can't foresee if it's not there. It's not like it will cause errors in the front either way. :)

ranta (approved):

PR #92, opened 2022-10-11
backend: apartment: implement unconfirmed max_prices

toikarin-vincit:
e71d527: backend: check migration issues: use `--dry-run`
 - this way no migration files are generated on this check

---

f7b61cd: backend: apartment: implement unconfirmed max_prices
 - implemented queries to fetch the data
 - updated OpenAPI
 - added tests

---

533fe25: backend: rename pre2005 indices
 - pre2005 renamed to 2005equal100
 - 2005equal100 is used for apartments constructed *after* 2011, not
   before

---


backend/hitas/views/apartment.py
```diff
@@ -366,38 +427,73 @@ class ApartmentViewSet(HitasModelViewSet):
...
+        ROUND(
+            a.debt_free_purchase_price * current_{table}.value / NULLIF(original_{table}.value, 0)
+        ) AS max_price_{table}
+    FROM hitas_apartment AS a
+    LEFT JOIN hitas_{table} AS original_{table} ON
+        a.completion_date {comparison} '2005-01-01' AND original_{table}.month = DATE_TRUNC('month', a.completion_date)
+    LEFT JOIN hitas_{table} AS current_{table} ON
+        a.completion_date {comparison} '2005-01-01' AND current_{table}.month = DATE_TRUNC('month', NOW())
+    WHERE a.id = hitas_apartment.id
+"""
```
ranta, line 445: Is it not possible to do this with Django ORM, or is there some benefit we gain from doing this with raw SQL?

backend/hitas/views/apartment.py
```diff
@@ -366,38 +427,73 @@ class ApartmentViewSet(HitasModelViewSet):
...
+                    "cpi_pre2005": self.select_index("constructionpriceindexpre2005", pre2005=True),
+                    "mpi": self.select_index("marketpriceindex", pre2005=False),
+                    "mpi_pre2005": self.select_index("marketpriceindexpre2005", pre2005=True),
+                    "sapc": """
+    SELECT ROUND(a.surface_area * sapc.value)
+    FROM hitas_apartment AS a
+    LEFT JOIN hitas_surfaceareapriceceiling AS sapc
+        ON sapc.month = DATE_TRUNC('month', NOW())
+    WHERE a.id = hitas_apartment.id
+""",
```
ranta, line 494: Is it not possible to do this with Django ORM, or is there some benefit we gain from doing this with raw SQL?

backend/hitas/models/indices.py
```diff
@@ -22,27 +22,27 @@ class Meta:
...
+        verbose_name_plural = _("Market price indices for apartment constructed before January 2011")
 
 
-class MarketPriceIndexPre2005(AbstractIndex):
+class MarketPriceIndex2005Equal100(AbstractIndex):
     class Meta:
-        verbose_name = _("Market price index for apartments constructed before January 2005")
-        verbose_name_plural = _("Market price indices for apartment constructed before January 2005")
+        verbose_name = _("Market price index for apartments constructed in January 2011 onwards")
+        verbose_name_plural = _("Market price indices for apartment constructed in January 2011 onwards")
```
ranta, line 32: Seems like a typo here 🤔 
`Market price index for apartments constructed in January 2011 __and__* onwards`?

backend/hitas/models/__init__.py
```diff
@@ -14,9 +14,9 @@
 )
 from hitas.models.indices import (
     ConstructionPriceIndex,
-    ConstructionPriceIndexPre2005,
+    ConstructionPriceIndex2005Equal100,
     MarketPriceIndex,
-    MarketPriceIndexPre2005,
+    MarketPriceIndex2005Equal100,
```
ranta, line 19: I don't really understand, what's the reasoning behind the naming change? (More specifically the `Equal100` part)

backend/hitas/tests/apis/test_api_apartment.py
```diff
@@ -308,6 +348,98 @@ def test__api__apartment__retrieve(api_client: HitasAPIClient):
...
+
+
+@pytest.mark.django_db
+def test__api__apartment__retrieve__pre_2005__current_month_indices_missing(api_client: HitasAPIClient):
+    _test_max_prices(api_client, pre_2005=True, create_completion_indices=True, create_current_indices=False)
+
+
+@pytest.mark.django_db
+def test__api__apartment__retrieve__pre_2005__indices_missing(api_client: HitasAPIClient):
+    _test_max_prices(api_client, pre_2005=True, create_completion_indices=False, create_current_indices=False)
```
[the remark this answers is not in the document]
toikarin-vincit: I do agree. I opted for this approach as I wanted to have these important (money/calculation related) tests as readable as possible and list of arguments did not feel like a best approach. Did not try the alternative though

backend/hitas/views/apartment.py
toikarin-vincit: Unfortunately I could not find a way to do this. There's no way to to custom join with django ORM without using `extra`. Join here requires two comparisons and one of them uses `DATE_TRUNC`. I even tried to implement custom `ForeignObject` implementation. Maybe there's a way to do this with a custom implementation but that would require quite a bit of a deep dive into django's ORM's internals

backend/hitas/views/apartment.py
toikarin-vincit: see above

backend/hitas/models/__init__.py
toikarin-vincit: There's two indices for market price index:
- `Markkinahintaindeksi` is used for apartments constructed before *2011* (https://www.hel.fi/static/kv/asunto-osasto/hitas-markkinahintaindeksi.pdf)
 - `Markkinahintaindeksi 2005=100` is used for apartments constructed *2011* onwards (https://www.hel.fi/static/kv/asunto-osasto/hitas-indeksit-2005-100.pdf)
 
 The `pre 2005` part is wrong. It could be a `pre 2011` but looking at the https://www.hel.fi/static/kv/asunto-osasto/hitas-indeksit-2005-100.pdf the `2005-100` is part of the filename and it's named `Markkinahintaindeksi 2005=100` (meaning index was 100 in 2005). I did not want to invent a new name, I believe that would be more confusing :)

  -

backend/hitas/views/apartment.py
ranta: Lets leave it then, no need to waste time in trying solve what's already solved 😄

backend/hitas/models/__init__.py
ranta: All right, that part of the naming scheme wasn't clear from the commits, good to know :)

backend/hitas/models/indices.py
toikarin-vincit: I don't think `and` is necessary here.  [Few examples](https://en.bab.la/sentences/english/onwards)

PR #94, opened 2022-10-12
backend: Fix HousingCompany empty string to be a valid business id value

ranta:
- The empty string is converted to None when it's being validated

Requires [22ffb8a](https://github.com/City-of-Helsinki/hitas/pull/93/commits/22ffb8a4ca2ae548e83878066503ecf569268a85) from #93 to be fully functional

refs. HT-30 Comments. Fixes to address the comments had been made already, but this one issue seems to have still slipped through

### DIFF
--- a/backend/hitas/views/housing_company.py
+++ b/backend/hitas/views/housing_company.py
@@ -153,6 +153,8 @@ class HousingCompanyDetailSerializer(EnumSupportSerializerMixin, HitasModelSeria
         return instance
 
     def validate_business_id(self, value):
+        if value == "":
+            value = None
         validate_business_id(value)
         return value
 


backend/hitas/views/housing_company.py
```diff
@@ -153,6 +153,8 @@ def update(self, instance: HousingCompany, validated_data: dict[str, Any]):
         return instance
 
     def validate_business_id(self, value):
+        if value == "":
+            return None
```
toikarin-vincit, line 157: is there a reason to convert `""` to `None`? If not, updating `validate_business_id` seems more reasonable

ranta: This way we allow the frontend to pass `""` as a valid value instead of `null`, and keep our data cleaner (Either a valid business id value or `null`)

Other option would be to allow all `""` `null` and a valid business id, and change the field to accept blank values (currently it doesn't accept them)

toikarin-vincit: ok, so there's a reason to convert `""` to `None`. For me the confusing part here is the different handling of `None` and `""`:

 - pass `""` -> return `None`
 - pass `None` -> call `validate_business_id(None)` -> return `None`
 - anything else `validate_business_id(value)` -> `raise` or `return <value>`
 
Maybe this would be bit more clear:

```
if value == "":
    value = None
validate_business_id(value)
return value
```

This way it's more clear that `""` is converted to `None` on purpose and same logic applies to all values.

PR #96, opened 2022-10-20
frontend: Fix NumberInputField not accepting values

ranta:
The passed value was never actually used, oops 😅

### DIFF
--- a/frontend/src/common/components/formInputField/FormNumberInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormNumberInputField.tsx
@@ -19,7 +19,7 @@ export default function FormNumberInputField({
 }: FormNumberInputFieldProps): JSX.Element {
     const formatNumber = (v: string): number | null => {
         if (v === "") return null;
-        return Number(Number().toFixed(fractionDigits ? fractionDigits : 0));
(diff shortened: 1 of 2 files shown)


frontend/src/common/components/formInputField/FormNumberInputField.tsx
```diff
@@ -19,7 +19,7 @@ export default function FormNumberInputField({
 }: FormNumberInputFieldProps): JSX.Element {
     const formatNumber = (v: string): number | null => {
         if (v === "") return null;
-        return Number(Number().toFixed(fractionDigits ? fractionDigits : 0));
+        return Number(Number(v).toFixed(fractionDigits ? fractionDigits : 0));
```
toikarin-vincit, line 22: what's the expected output if `fractionDigits` is `2` and `v` is `2.10`? `2.10` or `2.1`? I would expect the former and this returns the latter (because of the surrounding `Number()` :thinking:

toikarin-vincit (approved):

frontend/src/common/components/formInputField/FormNumberInputField.tsx
ranta: Added improved formatting for the field and noticed there were some other unwanted things with and fixed those as well :)

PR #99, opened 2022-10-24
Codes index creation

vincit-matu:
Adds Codes (w/ indices [list,create,edit])

- also refactors the max prices to be dynamic, as this was so dependant on correct indices to work correctly. Thus quite untestable as a completely separate branch

### DIFF
--- a/frontend/src/features/codes/Codes.tsx
+++ b/frontend/src/features/codes/Codes.tsx
@@ -1,11 +1,10 @@
-import React, {useState} from "react";
+import React from "react";
 
 import {Tabs} from "hds-react";
 
 import IndicesList from "./IndicesList";
 
 const Codes = (): JSX.Element => {
-    const [filterParams, setFilterParams] = useState({name: ""});
     return (
         <div className="view--codes">
             <h1 className="main-heading">Koodisto</h1>
@@ -16,20 +15,17 @@ const Codes = (): JSX.Element => {
                     <Tabs.Tab>Laskentasäännöt</Tabs.Tab>
                     <Tabs.Tab>Rahoitusmuodot</Tabs.Tab>
                 </Tabs.TabList>
-                <Tabs.TabPanel className="view--codes__tab--indices">
-                    <IndicesList
-                        filterParams={filterParams}
-                        setFilterParams={setFilterParams}
-                    />
+                <Tabs.TabPanel className={"view--codes__tab--indices"}>
+                    <IndicesList />
                 </Tabs.TabPanel>
-                <Tabs.TabPanel>
-                    <h2>Postinumerot</h2>
+                <Tabs.TabPanel className={"view--codes__tab--postalcodes"}>
+                    <h1>Postinumerot</h1>
                 </Tabs.TabPanel>
-                <Tabs.TabPanel>
-                    <h2>Laskentasäännöt</h2>
+                <Tabs.TabPanel className={"view--codes__tab--rules"}>
+                    <h1>Laskentasäännöt</h1>
                 </Tabs.TabPanel>
-                <Tabs.TabPanel>
-                    <h2>Rahoitusmuodot</h2>
+                <Tabs.TabPanel className={"view--codes__tab--fundingmodes"}>
+                    <h1>Rahoitusmuodot</h1>
                 </Tabs.TabPanel>
             </Tabs>
         </div>
--- a/frontend/src/features/codes/Codes.tsx
+++ b/frontend/src/features/codes/Codes.tsx
@@ -1,8 +1,13 @@
-import React from "react";
+import React, {useState} from "react";
 
-import {Tabs} from "hds-react";
+import {LoadingSpinner, Tabs} from "hds-react";
 
-const Codes = () => {
+import {useGetIndicesQuery} from "../../app/services";
+import {ListPageNumbers, QueryStateHandler} from "../../common/components";
+import {IIndex, IIndexResponse} from "../../common/models";
+
+const Codes = (): JSX.Element => {
+    const [filterParams] = useState({});
     return (
         <div className="view--codes">
             <h1 className="main-heading">Koodisto</h1>
@@ -14,7 +19,7 @@ const Codes = () => {
                     <Tabs.Tab>Rahoitusmuodot</Tabs.Tab>
                 </Tabs.TabList>
                 <Tabs.TabPanel>
-                    <h2>Indeksit</h2>
+                    <CodesResultList filterParams={filterParams} />
                 </Tabs.TabPanel>
                 <Tabs.TabPanel>
                     <h2>Postinumerot</h2>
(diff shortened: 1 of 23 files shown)


frontend/src/features/codes/Codes.tsx
```diff
@@ -0,0 +1,35 @@
...
+                <Tabs.TabPanel className={"view--codes__tab--indices"}>
+                    <IndicesList />
+                </Tabs.TabPanel>
+                <Tabs.TabPanel className={"view--codes__tab--postalcodes"}>
+                    <h1>Postinumerot</h1>
+                </Tabs.TabPanel>
+                <Tabs.TabPanel className={"view--codes__tab--rules"}>
+                    <h1>Laskentasäännöt</h1>
+                </Tabs.TabPanel>
+                <Tabs.TabPanel className={"view--codes__tab--fundingmodes"}>
```
ranta, line 27: ~fundingmodes~
FinancingMethods
😛

frontend/src/features/codes/IndicesList.tsx
```diff
@@ -0,0 +1,258 @@
...
+            hitasToast(
+                saveError ? "Indeksin tallennus epäonnistui" : "Indeksi tallennettu onnistuneesti",
+                saveError ? "error" : "success"
+            );
+        }
+    }, [isSaving, saveError, saveData]);
+    return isSaving ? (
+        <LoadingSpinner />
+    ) : (
+        <Dialog
```
ranta, line 209: `QueryStateHandler` could be used here to simplify a little bit? 🤔 

```
return (
    <QueryStateHandler isLoading={isSaving}>
        <Dialog...
```

frontend/src/features/codes/IndicesList.tsx
```diff
@@ -0,0 +1,258 @@
...
+                    variant={"secondary"}
+                >
+                    Peruuta
+                </Button>
+                <Button
+                    onClick={handleSaveIndex}
+                    theme={"black"}
+                >
+                    Tallenna
+                </Button>
```
ranta, line 252: Or the loading indicator could alternatively be moved here with:

```
<SaveButton
    onClick={handleSaveIndex}
    isLoading={isSaving}
/>
```

frontend/src/features/codes/IndicesList.tsx
```diff
@@ -0,0 +1,258 @@
...
+            </QueryStateHandler>
+            <></>
+        </div>
+    );
+};
+
+const IndicesList = (): JSX.Element => {
+    const [currentIndexType, setCurrentIndexType] = useState(indexTypes[0]);
+    const [editMonth, setEditMonth] = useState<string | null>(null);
+    const [editValue, setEditValue] = useState<number | null>(null);
```
ranta, line 127: Doesn't seem like these are actually used anywhere except to set them as `null`, so they can be removed

frontend/src/features/codes/IndicesList.tsx
```diff
@@ -0,0 +1,258 @@
...
+const EditIndexDialog = ({indexType, formData, setFormData, editDialogOpen, closeDialog}) => {
+    const [saveIndex, {data: saveData, error: saveError, isLoading: isSaving}] = useSaveIndexMutation();
+    const handleSaveIndex = () => {
+        saveIndex({
+            data: formData,
+            index: indexType.label,
+            month: formData.month,
+        });
+        closeDialog();
+    };
```
ranta, line 196: This doesn't handle error cases yet. If the user enters an invalid date the modal is closed and no error messages shown (Error message is shown on the modal but it's now hidden). Also no error toast notification is shown

frontend/src/features/codes/IndicesList.tsx
```diff
@@ -0,0 +1,258 @@
...
+        },
+    });
+    return (
+        <div className="results">
+            <QueryStateHandler
+                data={data}
+                error={error}
+                isLoading={isLoading}
+            >
+                {!isLoading && !error ? (
```
ranta, line 95: No need to check for errors or loading when using `QueryStateHandler` :)

ranta (changes requested):

toikarin-vincit (changes requested):
few observations:
 - code page shows 4 pages initially (`1, 2, 3, 4`). when clicking next the page count increases up seven (`1, 2, 3, 4, 5, 6, 7`). After 7 it shows 2-8. Is this intended?
 - months should come in descending order, do we have already a ticket for this?
 - page count 12 does not work as not all indices start from January. IMO paging these do not make that much sense, they should be browseable by year. Not needed tobe fixed here but ticket would be nice.
 - Names should be `Markkinahintaindeksi (Ennen 2005)` -> `Markkinahintaindeksi` and `Markkinahintaindeksi (2005 Eteenpäin)` -> `Markkinahintaindeksi 2005=100`. Same for rakennuskustannusindeksi. `Enimmäishintaindeksi` -> `Luovutushintaindeksi`
 - changing index keeps the same page number. this can be confusing as different indices have different amounts of data (jumping from page 3 of  `Markkinahintaindeksi 2005=100` to `Markkinahintaindeksi` jumps from 2007 -> 1980
 - max price index might get removed. no need to support it in the frontend for now. at least it shouldn't be the first in the list
 - saving does not work, tries to save to `/api/v1/[object Object]/<month>`
 - maybe a nice simple addition would be to add the current month to the saving dialog if a month is not selected :thinking:

frontend/src/features/codes/IndicesList.tsx
vincit-matu: They are. The first use is right on the next lines:
```
const initialSaveData: IIndex = {
        indexType: indexTypes[0].label,
        month: editMonth || "",
        value: editValue || null,
    };
```
and are the way through which the Dialog form fields are populated when editing an index, instead of creating a new one.

frontend/src/features/codes/IndicesList.tsx
vincit-matu: Implemented with requested component

vincit-matu:
- The paging is done by a HDS-component, so that's intended behavior. As I think this is the first listing where we have over five pages it's the first time I'm encountering this. I assume we're using the basic version, but there exists a truncated version for it, too, which I think might be good to use everywhere instead (https://hds.hel.fi/storybook/react/?path=/docs/components-pagination--with-truncation). I think we're going to have a lot of over five page listings.
- Agreed on the months in descending order. Should that be handled by the front?
- Regarding the page count, I think 12 is more logical than the arbitrary 10 since we're dealing with months/years. But you're right in that it doesn't work as a clean year selector. And I agree on the browsability by concrete year (and not periods of one year == 12 months) would offer a nicer user experience.
- Index names and order fixed as requested
- Changing the index now also resets the view to the first page
- Saving works for me and Eemeli. We checked that very issue was fixed after you and me diagnosed it in our huddle in the first place. :o
- Using the current month as a default value in index creation is a very good idea for usability. Implemented.

toikarin-vincit (changes requested):
is tickets already created for the open items?

also I'm not able to add decimals for some reason :thinking: Adding `.` does not work for the value field

ranta:
> is tickets already created for the open items?
> 
> also I'm not able to add decimals for some reason 🤔 Adding `.` does not work for the value field

The FormInputField should define the number of decimal places allowed (the default is zero), otherwise it will prevent decimals on the field

frontend/src/features/codes/IndicesList.tsx
ranta: What i mean is, the only value they have is `null`. It's their initial value and the only value they are set to 😅 

So initialSaveData would be simplified to:
```
const initialSaveData: IIndex = {
        indexType: indexTypes[0].label,
        month: null || "",
        value: null || null,
    };
```

frontend/src/features/codes/IndicesList.tsx
vincit-matu: This should be fixed now. I made the dialog only close on a successful save - otherwise it should now stay open and show the errors next to the respective input field.

vincit-matu:
> is tickets already created for the open items?

Yes. The tickets for open items are HT-302, HT-303 and HT-304.
 
> also I'm not able to add decimals for some reason 🤔 Adding `.` does not work for the value field

The decimal issue is fixed - however I think it might be a good idea to have the number input component accept two fractional digits as default? They are used at least in all inputs for monetary values, area values and index values - making a no-decimal number input the exception and not the other way around.

toikarin-vincit (approved):
My findings have been addressed so :rocket:

frontend/src/features/codes/IndicesList.tsx
vincit-matu: I am (finally) in agreement, after testing the page without the above mentioned values (at all) - and it worked just as expected. 😅 😅
Hence, they're removed as being completely extraneous. This also makes the `initData()` function unnecessary, as it used to set the above values to null, but now only sets the `formData` variable to `initialSaveData`. So we're now directly doing that (meaning `setFormData(initialSaveData)`) where we used to call `initData()` before.

PR #103, opened 2022-10-25
Max price calculations - missing index handling

toikarin-vincit:
1b283be: backend: better missing indices handling for max price calculations
 - previously 500 was returned and missing indices were not really
   handled. now `409 Conflict` is returned
 - added tests

---

18c53ee: backend: minor code style improvement on how to determinate max index

---

8f21c0f: backend: add global 'valid_until` for max price calculations
 - add validity for the "global" max price calculations so there's
   no need to dig it up from the selected index calculation
 - also correct surface area price ceiling validity, it requires
   +1 additional month

---

d8a03cf: backend: add `created` timestamp to max price calculation

---

### DIFF
--- a/backend/hitas/calculations/max_price.py
+++ b/backend/hitas/calculations/max_price.py
@@ -4,6 +4,7 @@ from typing import List, Optional
 
 from dateutil.relativedelta import relativedelta
 from django.db.models import Prefetch, Sum
+from django.utils import timezone
 
 from hitas.models import (
     Apartment,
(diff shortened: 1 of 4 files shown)


backend/hitas/calculations/max_price.py
```diff
@@ -103,6 +103,7 @@ def calculate_max_price(
             "market_price_index": market_price_index,
             "surface_area_price_ceiling": surface_area_price_ceiling,
         },
+        "created": datetime.datetime.now(),
```
ranta, line 106: I would use Django's `now` time instead
```
from django.utils import timezone

now = timezone.now()
```

toikarin-vincit: fixed

ranta (approved):

PR #105, opened 2022-10-26
Surface area price ceiling validity

toikarin-vincit:
---

3a3a889: backend: calculate surface area price ceiling validity
 - previously this was not implemented correctly, it was
   calculation_date + 1 month
 - surface area price ceiling validity calculation is now
   done programmatically when maximum price calculation is done
 - in older system this information was stored in index but
   as this follows simple rules, no need to store the value there
 - added tests when surface area price ceiling is the maximum value

---

4e45e88: backend: update dependencies

---

b188a5b: frontend: update dependencies

---


ranta (approved):
Great idea to simplify the validity dates for indices 🚀

PR #108, opened 2022-10-31
Changed indices names as per Saila's request

vincit-matu:
Simple name change for the indices.


toikarin-vincit (approved):
My proposal for the names was here https://github.com/City-of-Helsinki/hitas/pull/99#pullrequestreview-1152867676 which is AFAIK more official terms (see https://www.hel.fi/static/kv/asunto-osasto/hitas-indeksit-2005-100.pdf). That way the names would be the same also in frontend+backend

PR #119, opened 2022-11-04
github: run `docker build`s as part of CI

toikarin-vincit:


ranta (approved):
This is a nice addition to have, but does the CI now take significantly longer to complete?
Would running e.g. `yarn build` be faster and give the same results?

toikarin-vincit:
> This is a nice addition to have, but does the CI now take significantly longer to complete? Would running e.g. `yarn build` be faster and give the same results?

It's true this makes the build slower. Personally I haven't yet had problems GitHub builds taking too long so I opted for more thorough check. I slowness becomes a problem then we can later change this. :)

PR #120, opened 2022-11-04
Demo fixes

toikarin-vincit:
5bd2564: backend: return failure when max price calculation results a negative result
 - when apartment share of housing company loans is greater than
   the share's debt free price then the resulted maximum price
   would be a negative number. after this change error is raised
   and `409 Conflict` returned with `invalid_calculation_result`
   error when `result <= 0`

---

972a6b9: backend: oracle migration: don't migrate indices with value `0`
 - indices contain values to the future with value `0`. we can
   better handle adding new indices so just ignore those

---

0f6ac09: backend: return indices in month's descending order

---


ranta (approved):
🚀 🚀

PR #122, opened 2022-11-07
max price calculation - loan date

toikarin-vincit:
# Hitas Pull Request

# Description

fa3a76a: backend: implement apartment's share of housing company loans
 - when creating a maximum price calculation it's now possible
   to provide a `apartment_share_of_housing_company_loans_date`.
   It's included into the calculation as a extra information
   (it does not affect the calculation in any way). When the date
   is now provided, current date is used instead.

---

8ca4df1:: backend: increment JSON version
 - after adding the `aparment_share_of_housing_company_loans_date` in
   the previous commit the JSON format changed. This commit hides the
   old calculations and updates the JSON format from `1` -> `2`.

---

## Pull request checklist

  - **Backend**

## Test plan

1. `POST ../maximum-prices` and check the returned json contains `calculations.construction_price_index.calculation_variables.apartment_share_of_housing_company_loans_date` and `calculations.market_price_index.calculation_variables.apartment_share_of_housing_company_loans_date` as current date.
2. Do the same while providing `{"apartment_share_of_housing_company_loans_date": "<date>"}` payload and check the response returned the provided value.
3. Create (verified) calculations with the current master version and check those not exposed in the APIs anymore (apartment's confirmed price is missing) and confirming / retrieving old calculations return 404


backend/hitas/tests/apis/test_api_apartment.py
```diff
@@ -398,6 +398,27 @@ def test__api__apartment__retrieve__migrated_max_price(api_client: HitasAPIClien
...
+
+    response = api_client.get(
+        reverse(
+            "hitas:apartment-detail",
+            args=[ap.housing_company.uuid.hex, ap.uuid.hex],
+        )
+    )
+
+    assert response.status_code == status.HTTP_200_OK, response.json()
+    assert response.json()["prices"]["maximum_prices"]["confirmed"] is None
```
ranta, line 419: I think the confirmed price should be shown even when the json version is old as it doesn't make the confirmed price any less valid.

toikarin-vincit: excellent point. I added a commit to address this and treat the old versions same way as migrated are.

PR #124, opened 2022-11-08
PDF generation

ranta:
# Hitas Pull Request

# Description

Initial version of creating PDFs for maximum price calculations

## Jira tickets

refs. HT-211, HT-212

## Pull request checklist

  - **Frontend**
  - **Backend**

## Test plan

- Go to any apartment details page, create a confirmed maximum price calculation, if it doesn't already have one, and then click on the "Lataa enimmäishinta laskenta" button


frontend/src/features/apartment/ApartmentDetailsPage.tsx
```diff
@@ -81,6 +81,28 @@ const LoadedApartmentDetails = ({data}: {data: IApartmentDetails}): JSX.Element
     const unconfirmedPrices = isPre2011
         ? data.prices.maximum_prices.unconfirmed.pre_2011
         : data.prices.maximum_prices.unconfirmed.onwards_2011;
+
```
vincit-matu, line 84: Maybe it would be good to have this as a service, so it could be reused in other components easily.

frontend/src/features/apartment/ApartmentDetailsPage.tsx
```diff
@@ -127,8 +158,9 @@ const LoadedApartmentDetails = ({data}: {data: IApartmentDetails}): JSX.Element
                             theme="black"
                             size="small"
                             variant="secondary"
+                            onClick={handlePDFButtonClick}
                         >
-                            Lataa Hinta-arvio
+                            Lataa enimmäishinta laskenta
```
vincit-matu, line 163: This should be one word, methinks? (==enimmäishintalaskelma)

vincit-matu (approved):
Front seems legit! 🚀

backend/hitas/views/utils/pdf.py
```diff
@@ -0,0 +1,17 @@
+from io import BytesIO
+from typing import Any
+
+from django.http import HttpResponse
+from django.template.loader import get_template
+from rest_framework import exceptions
+from xhtml2pdf import pisa
+
+
+def render_to_pdf(template: str, context: dict[str, Any]) -> HttpResponse:
```
toikarin-vincit, line 10: small nitpick: IMO this does a bit more than renders to pdf, it creates a HttpResponse. I would make this helper return the `result.getvalue()` and create the HttpResponse in the calling view. Other option is to rename this to `pdf_response` or similar

backend/hitas/views/apartment_max_price.py
```diff
@@ -123,6 +128,28 @@ def update(self, request, *args, **kwargs):
             ampc.delete()
             return Response(status=HTTPStatus.NO_CONTENT)
 
+    @action(detail=True, methods=["GET"])
+    def download(self, request, **kwargs) -> HttpResponse:
+        with model_or_404(ApartmentMaximumPriceCalculation):
+            mpc = ApartmentMaximumPriceCalculation.objects.get(uuid=kwargs["pk"])
```
toikarin-vincit, line 134: Do we want to download a specific calculation as "confirmed max price calculation" or do we want to have a endpoint that returns the latest confirmed max price calculation? The difference being the calculation id does not have to be provided.

backend/hitas/views/apartment_max_price.py
```diff
@@ -123,6 +128,28 @@ def update(self, request, *args, **kwargs):
...
+
+            # Make sure the calculation is confirmed
+            if mpc.confirmed_at is None:
```
toikarin-vincit, line 137: how is unconfirmed max price downloaded?

backend/hitas/views/apartment_max_price.py
```diff
@@ -123,6 +128,28 @@ def update(self, request, *args, **kwargs):
...
+                    {
+                        "error": "not_confirmed",
+                        "message": "Maximum price calculation is not confirmed.",
+                        "reason": "Conflict",
+                        "status": 409,
+                    },
+                    status=HTTPStatus.CONFLICT,
+                )
+
+            context = {"mpc": mpc}
```
toikarin-vincit, line 148: maybe the context should use `max_price_calculation` as the key, not `mpc` :)

frontend/src/app/services.ts
```diff
@@ -41,6 +42,25 @@ export class Config {
...
+    if (!apartment.prices.maximum_prices.confirmed) {
+        hitasToast("Enimmäishintalaskelmaa ei ole olemassa", "error");
+        return;
+    }
+
+    const url = `${Config.api_url}/housing-companies/${apartment.links.housing_company.id}/apartments/${apartment.id}/maximum-prices/${apartment.prices.maximum_prices.confirmed.id}/download`;
+    const init = {headers: new Headers({Authorization: "Bearer " + Config.token})};
+    fetch(url, init).then((response) => {
+        response.blob().then((blob) => {
+            const filename = `Enimmäishintalaskelma ${apartment.address.street_address} ${apartment.address.stair} ${apartment.address.apartment_number}`;
```
toikarin-vincit, line 55: how about returning the name as part of the server response (with `Content-Disposition`)

backend/config/urls.py
```diff
@@ -16,5 +17,15 @@
         path("api-auth/", include("rest_framework.urls", namespace="rest_framework")),
     ]
 
+
+if settings.DEBUG:
```
toikarin-vincit, line 21: are these endpoints only for testing or what's the point of these?

backend/hitas/tests/apis/apartment_max_price/test_api_apartment_max_price_pdf.py
```diff
@@ -0,0 +1,51 @@
...
+        # Skip validating response, this prevents `UnicodeDecodeError: 'utf-8' codec can't decode byte...`
+        openapi_validate_response=False,
+    )
+    assert response.status_code == status.HTTP_200_OK, response.json()
+    assert response.get("content-type") == "application/pdf"
+
+
+@pytest.mark.django_db
+def test__api__apartment_max_price__retrieve__migrated(api_client: HitasAPIClient):
+    # TODO
```
toikarin-vincit, line 50: implement? this should return 404

backend/hitas/views/apartment_max_price.py
ranta: It's not yet, that will be done as a separate task from this

backend/config/urls.py
ranta: Yep, only for testing purposes. These will be removed at some later point.

backend/hitas/views/apartment_max_price.py
ranta: Lets go with this for now and change it up if it's needed, but thats a nice idea 😄

backend/hitas/views/apartment_max_price.py
toikarin-vincit: that's fine. just thinking we don't need to define a new error response here. This could just always assume the calculation is confirmed?

backend/config/urls.py
toikarin-vincit: few `FIXME`s would make this more clear :)

backend/hitas/views/apartment_max_price.py
ranta: As the PDF is going to end up the wording "confirmed" ("vahvistettu"), i'd rather just block generating PDFs from unconfirmed "more in depth" calculations. IIRC there is never going to be a case when a PDF is needed for this "more in depth" calculation that isn't confirmed. (Unconfirmed max price PDF values are generated similarly to how the unconfirmed prices are on Apartment details page)

backend/config/urls.py
ranta: Ended up removing these endpoints already, they are useful for testing but it should also be fine to test using the correct API endpoint directly.

backend/hitas/views/apartment_max_price.py
```diff
@@ -123,6 +126,36 @@ def update(self, request, *args, **kwargs):
...
+
+            # Migrated maximum price calculations are missing necessary data necessary to create a PDF
+            if mpc.json_version is None:
+                raise HitasModelNotFound(model=ApartmentMaximumPriceCalculation)
+
+            # Make sure the calculation is confirmed
+            if mpc.confirmed_at is None:
+                return Response(
+                    {
+                        "error": "not_confirmed",
```
toikarin-vincit, line 142: this error should be added to OpenAPI but as this is getting removed(?) it does not matter that much :thinking:

backend/hitas/views/apartment_max_price.py
```diff
@@ -123,6 +126,36 @@ def update(self, request, *args, **kwargs):
             ampc.delete()
             return Response(status=HTTPStatus.NO_CONTENT)
 
+    @action(detail=True, methods=["GET"])
+    def download(self, request, **kwargs) -> HttpResponse:
+        with model_or_404(ApartmentMaximumPriceCalculation):
+            mpc = ApartmentMaximumPriceCalculation.objects.get(uuid=kwargs["pk"])
+
+            # Migrated maximum price calculations are missing necessary data necessary to create a PDF
+            if mpc.json_version is None:
```
toikarin-vincit, line 135: simpler would be to pass this with `get` :)

toikarin-vincit (approved):
Added few comments that can be addressed later too. This LGTM for now and enables an important feature. Great job! :+1:

backend/hitas/views/apartment_max_price.py
ranta: This is getting removed? 🤔 
Or replaced with another error later on?

PR #126, opened 2022-11-10
backend: add housing company improvements to max price calculations

toikarin-vincit:
# Hitas Pull Request

# Description

 - 2011 after constructed apartment's do not have any apartment improvements do not try to calculate them
 - move improvement calculation logic to own file. this is helpful in the future as we will have more complex improvement calculations for apartments constructed before 2011
 - response contains now a lot of information about the improvements
   - `calculations.construction_price_index.calculation_variables` now has `housing_company_improvements` as an object instead of a single number. the object has two elements, `items` and `summary`. `items` contains a list of improvements and `summary` contains a summary values across all improvements (such as total value, total accepted values etc)
  - `apartment_improvements` has changed to be an object too but will always be `null` for now.
  - same is true for `calculations.market_price_index.calculation_variables`

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Create a maximum price calculations (with improvements) and see the resulted calculation variables are sane :)

## Tickets

This pull request resolves all or part of the following ticket(s): HT-79

### DIFF
--- a/backend/hitas/calculations/__init__.py
+++ b/backend/hitas/calculations/__init__.py
@@ -1,1 +1,2 @@
+from hitas.calculations.improvement import calculate_housing_company_improvements_after_2010
 from hitas.calculations.max_price import calculate_max_price
--- a/backend/hitas/calculations/exceptions.py
+++ b/backend/hitas/calculations/exceptions.py
@@ -1,0 +1,6 @@
+class IndexMissingException(Exception):
+    pass
+
+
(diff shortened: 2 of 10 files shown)


ranta (approved):
Not sure I completely understand everything but looks fine 🚀 

If possible at all, I'd love it if changes were made in smaller chunks in the future, as that makes reviewing a bit easier :)

PR #127, opened 2022-11-10
Unconfirmed max price pdf

ranta:
# Hitas Pull Request

# Description

- Add the ability to generate PDF documents for apartment unconfirmed maximum price calculations.
- Contains some PDF styling, and the ability to download generated PDF from the frontend.
- Does not yet contain the ability to enter additional information (lisätietoja), but that will be added in another PR

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

Go to apartment details page and click on the "Lataa Hinta-arvio" button

## Tickets

This pull request resolves all or part of the following ticket(s): HT-218, HT-219, HT-268, HT-269

image

image


backend/hitas/views/apartment.py
```diff
@@ -615,3 +620,34 @@ def get_list_queryset(self):
             )
             .order_by("id")
         )
+
+    @action(detail=True, methods=["GET"])
+    def unconfirmed_prices_pdf(self, request, **kwargs) -> Union[HttpResponse, Response]:
+        apartment = self.get_object()
+        apartment_data = ApartmentDetailSerializer(apartment).data
```
toikarin-vincit, line 627: first instinct says this fetches a quite a bit of extra information that's not really needed here and gets discarded

backend/hitas/templates/unconfirmed_maximum_price.jinja
```diff
@@ -0,0 +1,113 @@
+{% extends "components/base.jinja" %}
```
toikarin-vincit, line 1: I would love to see some kind of tests to these. While testing PDFs is far from optional, these templates contain a bit of logic and tests would help not to break these. These are currently untested (in sense of the content that they have).

There's many ways to test these but one naive and fast wound be something like:
```
test__render_unconfirmed_max_price__surface_are_price_ceiling():
   context = {...}
   html = render_to_pdf("unconfirmed_maximum_price.jinja", context)
   assert "rajahinnan perusteella" in html
  render
test_            {% if prices.construction_price_index.value > prices.market_price_index.value %}
                <td>Hitas-huoneiston velaton enimmäishinta rakennuskustannusindeksillä laskettuna, euroa</td>
```

backend/hitas/templates/unconfirmed_maximum_price.jinja
```diff
@@ -0,0 +1,113 @@
...
+        </p>
+    {% endif %}
+
+    <b>Hitas-huoneistonne velaton enimmäishinta on
+        {% if prices.construction_price_index.maximum %}
+            {{ prices.construction_price_index.value|intcomma }} euroa rakennuskustannusindeksillä laskettuna.
+        {% elif prices.market_price_index.maximum %}
+            {{ prices.market_price_index.value|intcomma }} euroa markkinahintaindeksillä laskettuna.
+        {% else %}
+            {{ prices.surface_area_price_ceiling.value|intcomma }} euroa. Velaton enimmäishinta on laskettu voimassaolevan rajahinnan perusteella ({{ apartment.surface_area }} m² * ???? euroa).
```
toikarin-vincit, line 87: the `????` needs fixing. It's something that would be part of the **confirmed** calculation but not the unconfirmed. So it needs to be fetched separately for this rendering context

backend/hitas/views/apartment.py
```diff
@@ -615,3 +620,34 @@ def get_list_queryset(self):
...
+                status=HTTPStatus.CONFLICT,
+            )
+
+        filename = f"Hinta-arvio {apartment.address}.pdf"
+        context = {"title": filename, "apartment": apartment_data}
+        pdf = render_to_pdf("unconfirmed_maximum_price.jinja", context)
+        response = HttpResponse(pdf, content_type="application/pdf")
+        # Download file for user instead of opening it in the browser
+        response.headers["Content-Disposition"] = f"attachment; filename={filename}"
+        response["Access-Control-Expose-Headers"] = "Content-Disposition"
```
toikarin-vincit, line 652: I'd reorder this a bit :sweat_smile: 

```python
        context = {"title": filename, "apartment": apartment_data}
        pdf = render_to_pdf("unconfirmed_maximum_price.jinja", context)

        response = HttpResponse(pdf, content_type="application/pdf")
        response.headers["Content-Disposition"] = f"attachment; filename=Hinta-arvio {apartment.address}.pdf"
```

toikarin-vincit, line 652: this should be set with `CORS_EXPOSE_HEADERS`? (https://github.com/adamchainz/django-cors-headers#cors_expose_headers-sequencestr)

toikarin-vincit (commented):
The downloaded PDF works and looks good. Some technicalities to be addressed

backend/hitas/templates/unconfirmed_maximum_price.jinja
ranta: The `????` part is the index value for _Rajaneliöhinta_, which is not included in the unconfirmed calculations, so we need to get it separately yes. Left this here as a TODO to fix a bit later :)

backend/hitas/tests/apis/test_api_apartment.py
```diff
@@ -487,47 +531,19 @@ def _test_max_prices(
...
-    )
+    # Validate fetching PDF report of unconfirmed prices
 
     response = api_client.get(
-        reverse(
-            "hitas:apartment-detail",
-            args=[ap.housing_company.uuid.hex, ap.uuid.hex],
-        )
+        reverse("hitas:apartment-detail", args=[ap.housing_company.uuid.hex, ap.uuid.hex]) + "/unconfirmed_prices_pdf",
+        # TODO: Write a validator that can handle pdf return types.
```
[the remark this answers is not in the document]
ranta: This is the same issue as with this: https://github.com/City-of-Helsinki/hitas/blob/master/backend/hitas/tests/apis/apartment_max_price/test_api_apartment_max_price_pdf.py#L41

OpenAPI is trying to decode the PDF response as utf-8 and fails, so its skipped instead.
(Here: https://github.com/p1c2u/openapi-core/blob/master/openapi_core/contrib/django/responses.py#L13)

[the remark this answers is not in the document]
toikarin-vincit: https://github.com/City-of-Helsinki/hitas/pull/127/commits/7f197d7d0874901e9fb4a7489acd94ef6fef0145 :-)

backend/hitas/views/apartment.py
ranta: You're correct, but the other way around this would be to write a serializer used only for fetching the necessary data for this, which I see as a bit too preemptive optimisation, so I went with using our already preexisting code to keep things simpler :)

vincit-matu (approved):

PR #128, opened 2022-11-10
Apartment share loans date

ranta:
# Hitas Pull Request

# Description

- Use correct field name on confirmed maximum price page for the field 'Yhtiölainaosuuden päivämäärä' in frontend.
- Format confirmed price dates 

## Pull request checklist

- **Frontend**

## Test plan

Set a date and verify that the field is correctly filled in the json on the pdf when downloading confirmed max price pdf

## Tickets

This pull request resolves all or part of the following ticket(s): HT-318

### DIFF
--- a/frontend/src/features/apartment/ApartmentMaxPricePage.tsx
+++ b/frontend/src/features/apartment/ApartmentMaxPricePage.tsx
@@ -193,6 +193,7 @@ const LoadedApartmentMaxPrice = ({apartment}: {apartment: IApartmentDetails}): J
                             formData={formData}
                             setFormData={setFormData}
                             error={error}
+                            maxDate={new Date()}
                         />
                         <FormInputField
                             inputType="date"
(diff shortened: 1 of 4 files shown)


frontend/src/features/apartment/ApartmentMaxPricePage.tsx
```diff
@@ -188,8 +189,7 @@ const LoadedApartmentMaxPrice = ({apartment}: {apartment: IApartmentDetails}): J
                         <FormInputField
                             inputType="date"
                             label="Yhtiölainaosuuden päivämäärä"
-                            fieldPath="housing_company_loans_date"
-                            required
+                            fieldPath="apartment_share_of_housing_company_loans_date"
```
toikarin-vincit, line 192: how easy it is to prevent selecting dates from the future? This would eliminate the `Field has to be less than or equal to current date.` error :thinking:

toikarin-vincit (approved):
Looking good. Just added one improvement comment that I accidentally faced when testing this.

vincit-matu (approved):

frontend/src/features/apartment/ApartmentMaxPricePage.tsx
ranta: Pretty eaasy, added a this :)

PR #130, opened 2022-11-11
Remove apartment

vincit-matu:
# Hitas Pull Request

# Description

Adds apartment removal feature. Introduces the ConfirmDialogModal component, which is used here and can be used to confirm any action via a dialogue modal.

## Pull request checklist

- **Frontend**

## Test plan

Navigate to any housing company page, and create a new apartment. Note, that there is no remove-button in this view when creating a new apartment. After it's created, click on "Muokkaa" to enter the creation/edit view once more - this time with a "Poista"-button visible as an option. Click it, and confirm your action in the modal which appears. You should be redirected back to the housing company page,  receive a toast notification stating that the operation succeeded.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-35

### DIFF
--- a/frontend/src/app/services.ts
+++ b/frontend/src/app/services.ts
@@ -252,6 +252,26 @@ const mutationApi = hitasApi.injectEndpoints({
                 return [];
             },
         }),
+        removeApartment: builder.mutation<
+            IApartmentDetails,
+            {
+                data: IApartmentWritable;
+                id?: string;
+                housingCompanyId: string;
+            }
+        >({
+            query: ({data, id, housingCompanyId}) => ({
+                url: `housing-companies/${housingCompanyId}/apartments${idOrBlank(id)}`,
+                method: "DELETE",
+                body: data,
+                headers: {"Content-type": "application/json; charset=UTF-8"},
+            }),
+            invalidatesTags: (result, error, arg) => [
+                {type: "Apartment", id: "LIST"},
+                {type: "Apartment", id: arg.id},
+                {type: "HousingCompany", id: arg.housingCompanyId},
+            ],
+        }),
         saveIndex: builder.mutation<IIndex, {data: IIndex; index: string; month: string}>({
             query: ({data, index, month}) => ({
                 url: `indices/${index}/${month}`,
@@ -285,6 +305,7 @@ export const {
     useCreateRealEstateMutation,
     useCreateBuildingMutation,
     useSaveApartmentMutation,
+    useRemoveApartmentMutation,
     useSaveApartmentMaximumPriceMutation,
     useSaveIndexMutation,
 } = mutationApi;
--- a/frontend/src/app/services.ts
+++ b/frontend/src/app/services.ts
@@ -255,15 +255,13 @@ const mutationApi = hitasApi.injectEndpoints({
         removeApartment: builder.mutation<
             IApartmentDetails,
             {
-                data?: IApartmentWritable;
                 id: string | undefined;
                 housingCompanyId: string;
             }
         >({
-            query: ({data, id, housingCompanyId}) => ({
+            query: ({id, housingCompanyId}) => ({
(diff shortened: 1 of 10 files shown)


frontend/src/common/components/ConfirmDialogModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+            aria-labelledby="confirm-modal"
+            isOpen={isVisible}
+            close={() => setIsVisible(false)}
+            boxShadow
+        >
+            <Dialog.Header
+                id="confirmation-modal__header"
+                title="Vahvista toiminto"
+            />
+            {isConfirmed && error ? (
```
ranta, line 95: I don't see the point in this variable. If error exists, it means that the 'confirm' button has been already pressed 🤔 
This could be removed without it making any changes to functionality

ranta (changes requested):

frontend/src/styles/components/_DialogModule.sass
```diff
@@ -6,3 +6,6 @@
 
   [class^="Table-module_actionContainer"]
     display: none
+
+  .row--buttons
+    margin-top: $spacing-layout-m
```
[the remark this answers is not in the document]
vincit-matu: Well spotted! 👍 I made `margin-top: $spacing-layout-m` the general rule, and thus specified it in _general.sass where the rest of the `.row--buttons` rules are located.

frontend/src/features/apartment/ApartmentCreatePage.tsx
```diff
@@ -77,7 +95,7 @@ const ApartmentCreatePage = () => {
                       end: null,
                   },
                   address: {
-                      street_address: "",
+                      street_address: (buildingOptions.length === 1 && buildingOptions[0]?.value) || "",
```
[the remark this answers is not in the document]
vincit-matu: 👍

frontend/src/common/components/ConfirmDialogModal.tsx
vincit-matu: Yup - you're right. It's a remnant from when I was thinking that the confirmation window would itself show a message after a successful confirmed action. But as the toast notification is used for this, it is unnecessary.

frontend/src/app/services.ts
```diff
@@ -240,6 +240,26 @@ const mutationApi = hitasApi.injectEndpoints({
...
+            {
+                data?: IApartmentWritable;
+                id: string | undefined;
+                housingCompanyId: string;
+            }
+        >({
+            query: ({data, id, housingCompanyId}) => ({
+                url: `housing-companies/${housingCompanyId}/apartments/${id}`,
+                method: "DELETE",
+                body: data,
```
ranta, line 254: This line can be completely removed from this request here, few lines above and where this function is called. `data` doesn't need to be sent to the API at all when removing an object :)

frontend/src/styles/components/_DialogModule.sass
```diff
@@ -5,4 +5,4 @@
     text-transform: capitalize
 
   [class^="Table-module_actionContainer"]
-    display: none
+    display: none
```
ranta, line 8: Two random spaces here 😅

frontend/src/app/services.ts
vincit-matu: Ah - I see! I understood from the earlier comment that it was supposed to be a non-required part of the request. Now removed entirely from both the mutation function in `services.js` and the function call in `ApartmentCreatePage.tsx`

frontend/src/styles/components/_DialogModule.sass
vincit-matu: Haha - it seems that's what you get when you try out a different editor. Tested VSCode, and it didn't have a eslint/prettier fix-on-save enabled. :D

ranta (approved):
LGTM 🚀

PR #131, opened 2022-11-15
Maximum price calculation additional info field

ranta:
# Hitas Pull Request

# Description

- Add `additional_field` to confirmed and unconfirmed maximum price PDFs.
- For confirmed maximum price, it's entered when creating a calculation.
- For unconfirmed (estimate) it's entered before downloading the PDF.
  - The old button now opens a modal, with 'Lisätieto'-field. The value is sent to the API and the text is inserted in the PDF when generating it.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Create a confirmed price calculation and make sure that the json on the pdf contains `additional_info` and the text you entered
- Download an unconfirmed price PDF, and enter some text in the model, make sure the PDF contains the text you entered.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-329


backend/openapi.yaml
```diff
@@ -926,6 +930,19 @@ paths:
           schema:
             type: string
             example: b477389cb3514fb1b444052a39bfb65d
+      requestBody:
+        required: true
```
toikarin-vincit, line 934: This could be optional?

backend/openapi.yaml
```diff
@@ -926,6 +930,19 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              description: TBD
```
toikarin-vincit, line 938: define :)

backend/openapi.yaml
```diff
@@ -926,6 +930,19 @@ paths:
...
+              type: object
+              additionalProperties: false
+              properties:
+                additional_info:
+                  description: Additional information to be displayed on the generated PDF
```
toikarin-vincit, line 943: `additional_info` is the correct translation but I'm thinking is a `notes` a better term (we already use that). `additional_info` sounds more like an object than a string. Other option would be to add "Additional **textual** information" to the description

frontend/src/features/apartment/ApartmentDetailsPage.tsx
```diff
@@ -76,16 +102,144 @@ const ConfirmedPrice = ({confirmed}) => {
...
+                    variant="secondary"
+                    theme="black"
+                    size="small"
+                >
+                    Sulje
+                </Button>
+                <Button
+                    theme="black"
+                    size="small"
+                    variant="secondary"
```
vincit-matu, line 149: This button shouldn't have the `variant="secondary"` attribute, as it's the main action button in the modal.

vincit-matu (changes requested):

frontend/src/features/apartment/ApartmentDetailsPage.tsx
ranta: Fixed :)

toikarin-vincit (approved):
LGTM :rocket:

vincit-matu (approved):

PR #136, opened 2022-11-17
Ownership add owner

vincit-matu:
# Hitas Pull Request

# Description

It's now possible to create a new owner if one doesn't already exist in the list, when creating or editing an apartment.

## Pull request checklist

- **Frontend**

## Test plan

Navigate to an apartment creation or edit page, and to the ownerships list. Click on the owner name to select a new owner. There is now a "Lisää omistaja" button which will take you to a form in which to create an owner. Once an owner is successfully created, the modal closes and the newly created owner is selected in the ownership.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-277

### DIFF
--- a/frontend/src/common/components/formInputField/FormOwnershipInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormOwnershipInputField.tsx
@@ -1,0 +1,241 @@
+import React, {useEffect, useState} from "react";
+
+import {Button, Dialog, IconCrossCircle, IconPlus, IconSearch, Table, TextInput} from "hds-react";
+import {useImmer} from "use-immer";
+
+import {useCreateOwnerMutation} from "../../../app/services";
+import {IOwner} from "../../models";
+import {hitasToast} from "../../utils";
+import QueryStateHandler from "../QueryStateHandler";
+import SaveButton from "../SaveButton";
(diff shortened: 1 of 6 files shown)


frontend/src/common/components/formInputField/FormOwnershipInputField.tsx
```diff
@@ -0,0 +1,241 @@
...
+                        isLoading={isLoading}
+                        onClick={confirmAction}
+                    />
+                )}
+            </Dialog.ActionButtons>
+        </>
+    );
+};
+
+export default function FormOwnershipInputField({
```
ranta, line 78: Not really a fan of each model having it's own component implemented this way. Here we have so much duplicate code vs `FormRelatedModelInputField`, but it's fine for now and let's refactor it into a more reusable component when we have a bigger need for it :)

ranta (approved):

frontend/src/common/components/formInputField/FormOwnershipInputField.tsx
vincit-matu: Agreed. A more generic component for adding items to a `FormRelatedModelInputField` might be a good idea.

PR #138, opened 2022-11-17
frontend: removes decimals from displayed unconfirmed max prices

vincit-matu:
# Hitas Pull Request

# Description

Removes the decimals from the displayed values for unconfirmed maximum prices in the apartment details view.

## Pull request checklist

- **Frontend**

## Test plan

Check out the apartment detail page. The different index maximum prices shown on the top left should not have any decimals.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-217 (comment from Saila)

### DIFF
--- a/frontend/src/features/apartment/ApartmentDetailsPage.tsx
+++ b/frontend/src/features/apartment/ApartmentDetailsPage.tsx
@@ -84,7 +84,7 @@ const UnconfirmedPriceRow = ({label, unconfirmedPrice}: UnconfirmedPriceRowProps
         <div className={`price${unconfirmedPrice.maximum ? " price--current-top" : ""}`}>
             <span className="basis">{label}</span>
             <span className="amount">
-                <span className="value">{formatMoney(unconfirmedPrice.value)}</span>
+                <span className="value">{formatMoney(Math.floor(unconfirmedPrice.value))}</span>
             </span>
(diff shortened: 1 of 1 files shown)


ranta (commented):
I'm thinking this change should maybe be done in backend (we need to change it in other places as well, e.g. PDFs), let's discuss this with the team.

miqu:
Yeah might be reasonable to not floor in everyplace. But if theres visible place where the decimals would be needed from the interface I think we would not need to duplicate the interface to get them for that one case. At the moment I dont see one so maybe just drop the decimals in the backend responce.

vincit-matu:
> I'm thinking this change should maybe be done in backend (we need to change it in other places as well, e.g. PDFs), let's discuss this with the team.

But I'm not doing a destructive operation here as I'm only affecting the _displayed_ value.. No matter how it's handled by the backend the displayed number won't show decimals. This works even if the value coming from the backend would already be an integer and it's not like one `Math.floor()` is an expensive operation to do in the clients end.. 😅 

> Yeah might be reasonable to not floor in everyplace.

The request is specifically to remove the decimals (I mean vs. asking to round to the nearest integer), and as this only affects the displayed value in one place I really don't see the harm.

ranta:
Replaced by #185

PR #141, opened 2022-11-18
Maximum Price Calculation fixes

ranta:
# Hitas Pull Request

# Description

- Don't calculate maximum prices with indices that have a `0` value
- Fix ordering for confirmed maximum prices (Correct calculation is now returned instead of the one with the highest price)

## Pull request checklist

- **Backend**

## Test plan

-

## Tickets

This pull request resolves all or part of the following ticket(s): None


backend/hitas/calculations/improvement.py
```diff
@@ -164,7 +164,12 @@ def calculate_improvement(
 ) -> ImprovementCalculationResult:
     deprecation_result = None
 
-    if index_check and (index_check.completion_date_index is None or index_check.calculation_date_index is None):
+    if index_check and (
+        index_check.completion_date_index is None
+        or index_check.completion_date_index == 0
```
toikarin-vincit, line 169: as we now have this https://github.com/City-of-Helsinki/hitas/blob/master/backend/hitas/oracle_migration/runner.py#L116-L118 it would be better to enforce index value must be greater than 0 instead of greater than or equal to zero. Allowing setting the index value to zero does not make sense anymore :thinking:

backend/hitas/views/indices.py
```diff
@@ -93,7 +93,7 @@ def get_object(self):
             return self.model_class(month=self.kwargs["month"])
 
     def get_queryset(self):
-        return self.model_class.objects.only("month", "value").all().order_by("-month")
+        return self.model_class.objects.filter(value__gt=0).only("month", "value").all().order_by("-month")
```
toikarin-vincit, line 96: no need for filter when value is constrained to be `>= 0`?

backend/hitas/calculations/improvement.py
toikarin-vincit: these checks can now be removed

backend/hitas/views/indices.py
ranta: My thinking was that there still may be old `0` value indices in the db. Instead of this i will add a migration to completely remove those.

toikarin-vincit: there might. I always forget that `validators` do not add the database constraints automatically :cry: It requires a manual
```
constraints = [
    models.CheckConstraint(name="index_value_positive", check=models.Q(value__gte=0)),
]
```

these would be nice to have to enforce consistency between code and the database

backend/hitas/calculations/improvement.py
ranta: 👍

toikarin-vincit (approved):
:rocket:

PR #144, opened 2022-11-21
backend: allow filtering housing companies by completion date

toikarin-vincit:
# Hitas Pull Request

# Description

 - query parameter is named `new_hitas`.
   - When this is given and set to `true`, only housing companies completed `2011-01-01` onwards are returned.
   - When this parameter is give nand set to `false` only housing companyes completed before `2011-01-01` are returned

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

Fetch `/api/v1/housing-companies` with `new_hitas=true`, `new_hitas=false` and without `new_hitas` and observe the results.

## Tickets

This pull request resolves all or part of the following ticket(s): n/a

### DIFF
--- a/backend/hitas/tests/apis/test_api_housing_company.py
+++ b/backend/hitas/tests/apis/test_api_housing_company.py
@@ -1,3 +1,4 @@
+import datetime
 from datetime import date
 from decimal import Decimal
 from typing import Any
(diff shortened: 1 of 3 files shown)


ranta (approved):
Nice 🚀 
Can you also create a Jira ticket to add this filter to the frontend too (If we don't already have one) :)

PR #145, opened 2022-11-23
Confirm calculation breakdown

vincit-matu:
# Hitas Pull Request

# Description

Adds a breakdown of the max price calculation to the confirmation dialog, which is shown when the user wants to save a calculation.

## Pull request checklist

- **Frontend**

## Test plan

Navigate to any apartment max price page, and click on "Laske". The dialog window should show the indices on the left, with the max price one being initially selected, and with a bold font (even if you select another index). Check all of the different indices by clicking on them. Rajaneliöhinta should have different contents from the other two indices.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-217

### DIFF
--- a/frontend/src/common/models.ts
+++ b/frontend/src/common/models.ts
@@ -279,20 +279,25 @@ export interface IApartmentWritable {
 
 // //  Maximum Price Fields
 
-interface ICalculation {
+export interface ICalculation {
     maximum_price: number;
     valid_until: string;
     maximum: boolean;
 }
 
-interface IIndexCalculationVariables {
+export interface IIndexCalculationVariables {
     calculation_variables: {
         acquisition_price: number;
         additional_work_during_construction: number;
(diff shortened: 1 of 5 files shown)


frontend/src/common/models.ts
```diff
@@ -279,20 +279,25 @@ export interface IApartmentWritable {
...
         basic_price: number;
         index_adjustment: number;
         apartment_improvements: number;
-        housing_company_improvements: number;
+        housing_company_improvements: {
+            summary: {
+                value: number;
+                value_for_apartment: number;
+            };
+        };
```
ranta, line 300: Missing some fields, but we can add them when they become necessary, the json is not fully done in the backend yet anyway :)

vincit-matu: Yep, my thoughts exactly.

PR #146, opened 2022-11-24
List indices by year

vincit-matu:
# Hitas Pull Request

# Description

Adds a year filter to the indices list, as the main way to navigate within the listing. Removes all pagination related components from the page, as they are now unnecessary. Also adds a dynamic minLength and maxLength to the FilterTextInputField component, so that the index filter only responds to values with four characters.

## Pull request checklist

- **Frontend**

## Test plan

Check out the indices list (== click on "Koodisto"). Type a year into the "Vuosi" filter input and watch magic happen!  🪄 🎉  

## Tickets

This pull request resolves all or part of the following ticket(s): HT-304


ranta (approved):
🦞

PR #155, opened 2022-12-07
Apartment type no longer required in creation, form error styles fixed

vincit-matu:
# Hitas Pull Request

# Description

Corrects the apartment creation required fields (== removes the `required` attribute from apartment type, otherwise all is as it should). Fixes styles so that the error messages for input fields are shown under the field in question.

## Pull request checklist

- **Frontend**

## Test plan

Go to creating a new apartment, and press "Tallenna" without filling in any fields. The apartment type still shows an error, but that is due to the API response from the backend.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-237


ranta (approved):

toikarin-vincit (approved):
updated test would be nice to have, otherwise LGTM :hippopotamus:

PR #158, opened 2022-12-08
backend: Display correct value for debt-free-price in SAPC Calculation

ranta:
# Hitas Pull Request

# Description

Previously the maximum price for surface area * calculation_date_value had debts already removed, now show the value without removing the debt part first.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

Create a new maximum price calculation with debt that uses surface area price ceiling and check out the PDF 

## Tickets

This pull request resolves all or part of the following ticket(s): HT-347


toikarin-vincit (approved):
:elephant:

PR #164, opened 2022-12-13
Max price breakdown

ranta:
# Hitas Pull Request

# Description

- Frontend and PDF breakdowns of maximum price calculations for all pre-2011 cases
- Some bugfixes/refactoring

## Pull request checklist

Check the boxes for each DoD items that has been completed:

- **Frontend**
    - [x] Changes have been tested
- **Backend**
    - [x] Changes have been tested
    - [x] Automatic tests has been added
    - [x] OpenAPI definitions have been updated
    - [ ] Database migrations will work in test environment
    - [ ] Oracle migration has been updated

## Test plan

- Create a new max price calculation on a pre-2011 apartment
  - Check out the new and amazing breakdowns for all different indices
- Confirm the max price calculation
  - Download the PDF for max price
    - Be amazed at how beautiful it is

## Tickets

This pull request resolves all or part of the following ticket(s): HT-315, HT-340, HT-353

PR #166, opened 2022-12-14
Removes extraneous line from the max price breakdown

vincit-matu:
# Hitas Pull Request

# Description

There was a line in the confirmed maximum price breakdown which had an error, and turns out was not needed at all as it proved to be unnecessary in its entirety (as per Saila's comment in HT-330). That line is removed here.

## Pull request checklist

- **Frontend**
    
## Test plan

Go to confirm a max price, and see that there is no line with "Huoneistokohtaiset parannukset".

## Tickets

This pull request resolves all or part of the following ticket(s): HT-330


ranta:
Can be closed, as this fix was included in #164

PR #168, opened 2022-12-15
Valid identifiers

toikarin-vincit:
# Hitas Pull Request

# Description
4dfc465: backend: minor test improvement
 - check for response json in `test__api__owner__create__invalid_data`

---

1870781: backend: add validate social security number function

---

20c80ee: backend: don't allow to change a valid ssn to an invalid one

---

26c0aa2: backend: oracle migration: skip duplicate owners with same ssn and name
 - removes duplicate persons with a same valid ssn and same name

---

a36a483: backend: fully implement `validate_business_id`
 - previous version did not check for the checksum, only that the format
   was correct
 - now checksum is calculated and previously valid business ids such as
   `1234567-8` are not valid anymore
 - added separate `check_business_id` which does not raise an exception
   but can be used to more gracefully check for a business id validness

---

f648668: backend: rename `validate_social_security_number` -> `check_social_security_number`
 - previous commit added `check_business_id` so align this with it
   (this does not raise an exception when the given social security
   number is invalid either)

---

82e8dca: backend: don't allow to change a valid business_id to an invalid one
 - rename `valid_ssn` to `valid_identifier`
 - check for valid business id when saving the object
 - prevent changing a valid business id to an invalid one
 - backend: oracle migration: skip duplicate owners with same business id and name
 - OwnerFactory now generates both ssn (90%) and business id (10%)

---

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Run the migration and check you're unable to find two owners with the same identifier and same name
- Updating a valid identifier to an invalid one (ssn or business id) is prohibited
- Updating a valid identifier to a another valid one (ssn or business id) is allowed
- Updating a invalid identifier to an invalid one (ssn or business id) is allowed
- Updating a invalid identifier to a valid one (ssn or business id) is allowed

## Tickets

This pull request resolves all or part of the following ticket(s): HT-180


backend/hitas/models/utils.py
```diff
@@ -49,3 +50,59 @@ def validate_building_id(value: Optional[str]) -> None:
...
+    date_str = ssn_match.group(1)
+    century_sign = ssn_match.group(2)
+    individual_number_str = ssn_match.group(3)
+    check_digit = ssn_match.group(4)
+
+    # Figure out the century from the century sign
+    match century_sign:
+        case "A" | "B" | "C" | "D" | "E" | "F":
+            century = 2000
+        case "-" | "Y" | "X" | "W" | "V" | "U":
```
ranta, line 117: TIL the century sign can be other than `A` or `-` 😄

ranta (approved):
Nice 🐘

backend/hitas/models/utils.py
toikarin-vincit: not yet, but those are valid starting from 2023 :) so this is future-proofed! :grin:

PR #170, opened 2022-12-19
Frontend CI

ranta:
# Hitas Pull Request

# Description

Add frontend style checking to Github CI

## Test plan

See if tests are run and pass :)


vincit-matu (approved):
It seems github is at least running the "check for prettier" in the CI pipeline - so I guess it's GTM.

PR #175, opened 2022-12-27
Spinner and headings

vincit-matu:
# Hitas Pull Request

# Description

Adds the Heading component, to function as a wrapper component for our header texts and buttons. Also restyles the loading spinner and the error page shown when there is no response from the backend.

## Pull request checklist

- **Frontend**

## Test plan

The Heading component is by design hard to notice, altho there is some vertical centerings which were off by a few pixels that are now fixed. Other than that all headings should look and work exactly as they did before. Implemented for all main headings and list headings.

To see the loading spinner for more than a few (micro)seconds use e.g. Redux DevTools browser extension to enable rewinding the view to a state where it has yet to receive a response from the API. The error screen on the other hand can be seen e.g. by running only the front while the backend is down.

## Tickets

This pull request resolves all or part of the following ticket(s):

### DIFF
--- a/frontend/src/common/components/QueryStateHandler.tsx
+++ b/frontend/src/common/components/QueryStateHandler.tsx
@@ -37,7 +37,6 @@ export default function QueryStateHandler({
             </p>
         );
     }
-    console.log(error);
     return error ? (
         errorComponent
     ) : isLoading ? (
--- a/frontend/src/common/components/QueryStateHandler.tsx
+++ b/frontend/src/common/components/QueryStateHandler.tsx
@@ -28,15 +28,24 @@ export default function QueryStateHandler({
     children,
 }: QueryLoadingProps): JSX.Element {
     // When loading or an error has occurred, show an appropriate message, otherwise return children
-
+    const errorMessage = error ? (error as {error: string}).error : "";
     if (errorComponent === undefined) {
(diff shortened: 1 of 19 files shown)


frontend/src/common/components/QueryStateHandler.tsx
```diff
@@ -28,15 +28,24 @@ export default function QueryStateHandler({
...
-        errorComponent = <p>Virhe</p>;
+        errorComponent = (
+            <p className="query-handler-error">
+                Ladattaessa tapahtui virhe 😞
+                {error !== undefined && <code className="query-error-message">{errorMessage as string}</code>}
+            </p>
+        );
     }
-
+    console.log(error);
```
ranta, line 40: console log can be removed

frontend/src/features/apartment/ApartmentCreatePage.tsx
```diff
@@ -241,12 +242,12 @@ const ApartmentCreatePage = () => {
 
     return (
         <div className="view--create view--set-apartment">
-            <h1 className="main-heading">
+            <Heading>
```
ranta, line 245: Looks like replacing Apartment heading wasn't as simple as this, it still needs some style fixes
image

ranta (changes requested):

ranta (approved):

PR #182, opened 2022-12-29
Add interest during construction

matti-lamppu:
# Hitas Pull Request

# Description

- Adds 6% or 14% interest to pre-2011 index calculations on the `/api/v1/housing-companies/{housing_company_id}/apartments/{apartment_id}` endpoint.
- Refactored old SQL strings to django ORM methods. New methods use subqueries instead of aliased joins, and they are much easier to read IMO.
    - In theory, subqueries might be slightly slower due to SQL compilers having better micro-optimization for joins, but I didn't measure any difference with debug-toolbar
- Added conditionals for adding thee interest based on the index calculation wanted.
- Added `select_related` for `financing_method` on the queryset for optimization since `old_hitas_ruleset` is needed to determine if interest should be used.
- Switched from `extra` to `annotations` since extra is [soft-deprecated](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#:~:text=Use%20this%20method%20as%20a%20last%20resort).
- Refactored surface area price ceiling from SQL strings to django ORM methods for consistency with the new index calculations. This also uses subqueries now (see the note above).
- Refactored tests to support `old_hitas_ruleset` as well as pre-2005 construction price index calculations

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [ ] Checked that unconfirmed maximum prices for apartments that finished construction before 2011/01/01 should now have interest included
    - [ ] For the market price index, it is the 6% interest
    - [ ] For the construction price index, it is the 6% interest if construction finished on or after 2005/01/01, and 14% if before
- [ ] Interest is only calculated if the financing method has the `old_hitas_ruleset` flag. Note that interest is not calculated for pre-2011 apartments if they use new rulesets.

## Tickets

HT-327 - Hinta-arvion laskenta muutokset - rakennusaikainen korko - backend

### DIFF
--- a/backend/hitas/tests/apis/test_api_apartment.py
+++ b/backend/hitas/tests/apis/test_api_apartment.py
@@ -494,6 +494,7 @@ def _test_max_prices(
         surface_area=50 if not null_values else None,
         interest_during_construction_6=1000,
         interest_during_construction_14=2000,
+        building__real_estate__housing_company__financing_method__old_hitas_ruleset=old_hitas_ruleset,
     )
 
     if create_completion_indices:
@@ -505,10 +506,6 @@ def _test_max_prices(
         cpi_factory.create(month=now, value=150)
         mpi_factory.create(month=now, value=250)
         SurfaceAreaPriceCeilingFactory.create(month=now, value=3000)
-
-    if old_hitas_ruleset:
-        ap.housing_company.financing_method.old_hitas_ruleset = old_hitas_ruleset
-        ap.housing_company.financing_method.save()
 
     # Validate apartment details has correct data returned
     response = api_client.get(
--- a/backend/hitas/tests/apis/test_api_apartment.py
+++ b/backend/hitas/tests/apis/test_api_apartment.py
@@ -40,7 +40,6 @@ from hitas.tests.factories.indices import (
 )
 from hitas.views.apartment import ApartmentDetailSerializer
 
-
 PRE_2005_DATE = datetime.date(2004, 12, 1)
 PRE_2011_DATE = datetime.date(2010, 12, 1)
 POST_2011_DATE = datetime.date(2011, 1, 1)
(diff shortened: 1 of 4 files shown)


backend/hitas/views/apartment.py
```diff
@@ -23,18 +23,18 @@
...
     SurfaceAreaPriceCeiling,
-    MarketPriceIndex,
-    MarketPriceIndex2005Equal100,
-    ConstructionPriceIndex,
-    ConstructionPriceIndex2005Equal100,
 )
 from hitas.models._base import HitasModelDecimalField
 from hitas.models.apartment import ApartmentMarketPriceImprovement, ApartmentState, DepreciationPercentage
-from hitas.utils import this_month, RoundWithPrecision
+from hitas.utils import RoundWithPrecision, this_month
```
ranta, line 37: And this is why the pre-commit hook is amazing (There would be no separate commit for this) 😄

backend/hitas/tests/apis/test_api_apartment.py
```diff
@@ -496,6 +499,10 @@ def _test_max_prices(
         mpi_factory.create(month=now, value=250)
         SurfaceAreaPriceCeilingFactory.create(month=now, value=3000)
 
+    if old_hitas_ruleset:
+        ap.housing_company.financing_method.old_hitas_ruleset = old_hitas_ruleset
```
ranta, line 510: This can be defined when creating the apartment like so:
```
ApartmentFactory.create(
    ...,
    building__real_estate__housing_company__financing_method__old_hitas_ruleset = old_hitas_ruleset
)
```

matti-lamppu: Done

ranta (approved):
LGTM 🦦

matti-lamppu:
image
feelsgood

PR #183, opened 2022-12-29
Prevent chaning owner identifier to existing one or creating duplicates

matti-lamppu:
# Hitas Pull Request

# Description

Prevent changing an owner's identifier to another existing identifier, or creating an owner with some existing owner's identifier.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

Tests can be done in the backend API (/api/v1/owners). Frontend does not correctly display these errors yet, but it can be verified that it no longer allows creating new owners with existing owner's identifiers.

- [ ] Owner cannot be created with some existing owner's identifier
- [ ] Owner cannot be modifier with some existing owner's identifier
- [ ] Owner can be otherwise modified when keeping its existing valid identifier in place
- [ ] Owner can be otherwise modified when keeping its existing invalid identifier in place

## Tickets

HT234 - Henkilön henkilötunnus ei ole uniikki

### DIFF
--- a/backend/hitas/tests/apis/test_api_owner.py
+++ b/backend/hitas/tests/apis/test_api_owner.py
@@ -153,7 +153,7 @@ def test__api__owner__create__invalid_data(api_client: HitasAPIClient, invalid_d
 
 
 @pytest.mark.django_db
-def test__api__owner__create__existing_identifier(api_client: HitasAPIClient):
+def test__api__owner__create__existing_valid_identifier(api_client: HitasAPIClient):
     data = get_owner_create_data()
     OwnerFactory.create(identifier=data["identifier"])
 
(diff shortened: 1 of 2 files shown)


backend/hitas/tests/apis/test_api_owner.py
```diff
@@ -224,6 +293,53 @@ def test__api__owner__update__invalid_identifier_to_invalid(api_client: HitasAPI
...
+    owner_2: Owner = OwnerFactory.create(identifier="bar")
+    data = {
+        "name": "Matti Meikäläinen",
+        "identifier": owner_2.identifier,
+        "email": "test@example.com",
+    }
+
+    url = reverse("hitas:owner-detail", kwargs={"uuid": owner_1.uuid.hex})
+    response = api_client.put(url, data=data, format="json")
+    assert response.status_code == status.HTTP_400_BAD_REQUEST, response.json()
```
ranta, line 328: This should not result in an error, since the identifier is not valid.
e.g. There may be cases where we dont have the full identifier of a person, only the birthday, and we should be able to save multiple people with the same birthday to the system 😄
Also now it's not possible to have multiple owners with `identifier = None`

ranta (changes requested):

backend/hitas/tests/apis/test_api_owner.py
matti-lamppu: This case is testing the update, so it's asserting that you shouldn't be able to change an existing owner's invalid identifier to some other owner's identifier, essentially merging them. Maybe that should be still allowed since no actual merging is happening. 

As for the creating part, I can make it not validate when adding new owners with invalid identifiers.

matti-lamppu: Changes made, new tests also added for null identifiers.

ranta: >Maybe that should be still allowed since no actual merging is happening?

I would say yes.

If it's not allowed, and multiple `Owner`'s have the same `identifier`, and one of them has only their name changed, a _'duplicate identifier error'_ is raised even though the identifier was unchanged

ranta (approved):
Nice array of tests, LGTM 🦛

backend/hitas/tests/apis/test_api_owner.py
matti-lamppu: `if value != self.instance.identifier` bypasses the checks if the identifier remains unchanged, so you could still change just the name and be fine. But I think allowing invalid identifiers to change to other invalid identifiers is fine based on what you said previously, so I'll make the merge.

PR #187, opened 2023-01-02
Front validate ssn

vincit-matu:
# Hitas Pull Request

# Description

Adds social security number validation to frontend. In practice it's a function in utils.js (for now, maybe better to have separate validators.js if they grow in number). Implemented now in the new owner creation dialogue, when selecting ownerships. The text-field shows a red border color as long as the value inside is not a valid SSN number.

If the number is not valid when saving, the form prompts the user for a confirmation to the saving event.

## Pull request checklist

- **Frontend**
    
## Test plan

Go to either creating or editing an apartment, and click on an ownership name to access the ownership selection. Click "Lisää omistaja" and feed in a valid SSN into the center input. It should get red borders from the first value change right until the value is a proper SSN.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-333


ranta (commented):
This behaves with SSNs as expected, but I think we need to be able to handle company identifiers _(Y-Tunnus)_ as well, even though there isn't 'supposed' to be more of them 🤔

vincit-matu:
> This behaves with SSNs as expected, but I think we need to be able to handle company identifiers _(Y-Tunnus)_ as well, even though there isn't 'supposed' to be more of them 🤔

This allows for saving Y-tunnus, too - you can save even without the identifier being a valid SSN.

ranta (approved):
🐼

PR #188, opened 2023-01-03
Add separate notes for improvements with mpi excess before and after 2010

matti-lamppu:
# Hitas Pull Request

# Description

Add separate notes for excess amounts applicable in market price index calculations for housing company improvements made before 2010/01/01 and after in apartments completed before 2011/01/01.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Calculate and download the **confirmed** max price calculations PDF for an apartment completed before 2011/01/01 with housing company improvements. 
    - Note that the highest price must be from the market price index calculation for the note to appear. One example is to use "Kimalaispolku 382, 00530, Helsinki" and set "Valmistumispäivä" to "2.6.2004"

## Tickets

This pull request resolves all or part of the following ticket(s): HT-379, HT-316

[HT-379]: https://helsinkisolutionoffice.atlassian.net/browse/HT-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/hitas/calculations/improvements/rules_pre_2011_mpi.py
```diff
@@ -219,6 +219,19 @@ class HousingCompanyImprovementCalculationResult(ApartmentImprovementCalculation
...
 @dataclass
 class HousingCompanyImprovementCalculationSummary(ApartmentImprovementCalculationSummary):
+    @dataclass
+    class ExcessSummary:
+        # Apartment's surface area
+        surface_area: Decimal
+        # Excess amount per square meter before 2010/01/01
+        value_per_square_meter_before_2010: Decimal
+        # Excess amount per square meter on or after 2010/01/01
+        value_per_square_meter_after_2010: Decimal
```
ranta, line 229: I would say no need to save these, as they are hardcoded values

matti-lamppu: I think we end up saving more if we save it for every improvement, and it's also easier to write the notes if the values are included in the excess data. Improvement already get's the `completion_date` so I can use that and add an `improvements_before_2010: bool` and `improvements_after_2010: bool` to the excess. Or I'll make the `value_per_square_meter_before_2010` and `value_per_square_meter_after_2010` nullable and determine based on that.

matti-lamppu: Made corresponding `value_per_square_meter` and `total` fields null if said excess is not applicable. This allows them to be hidden as well in the PDF.

backend/hitas/templates/components/improvements_table.jinja
```diff
@@ -147,6 +161,15 @@
                 <td class="improvement-cell text-right">{% if improvement.depreciation is not none %}{{ improvement.depreciation.amount|intcomma }}{% else %}-{% endif %}</td>
                 <td class="improvement-cell text-right">{{ improvement.accepted_value_for_housing_company|intcomma }}</td>
                 <td class="improvement-cell text-right">{{ improvement.accepted_value|intcomma }}</td>
+                {% if excess and excess.value_per_square_meter_before_2010 is not none and excess.value_per_square_meter_after_2010 is not none %}
+                <td class="improvement-cell text-right">
+                  {% if improvement.completion_date|date_before("2010-01-01") or excess.value_per_square_meter_before_2010 is none %}
+                    <small>1</small>
+                  {% else %}
+                    <small>2</small>
+                  {% endif %}
```
ranta, line 170: Nice and simple way to check for the date, instead of saving the excess for every improvement 👍

ranta (approved):
🧅

PR #189, opened 2023-01-04
Validate shares when creating/updating apartments

matti-lamppu:
# Hitas Pull Request

# Description

Shares in the same housing company should not overlap. Raise an error if this happens.

## Pull request checklist

- **Frontend**
- **Backend**
    - [ ] Database migrations will work in the test environment

## Test plan

- [x] Adding an apartment with overlapping shares raises an error about the shares
    - [x] Displays the overlapping share range
    - [x] Displays the apartment that the overlap is from
    - [x] Displays a single share if only the start or end of the range overlaps (both are inclusive)
        - [x] Displays only on the start field if the start overlaps
        - [x] Displays only on the end field if the end overlaps
- [x] Editing an apartment so that its shares now overlap with another building in the same housing company raises an error
    - [x] Displays the overlapping share range
    - [x] Displays the apartment that the overlap is from
    - [x] Displays a single share if only the start or end of the range overlaps (both are inclusive)
        - [x] Displays only on the start field if the start overlaps
        - [x] Displays only on the end field if the end overlaps

## Tickets

This pull request resolves all or part of the following ticket(s): HT-355

[HT-355]: https://helsinkisolutionoffice.atlassian.net/browse/HT-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/backend/hitas/views/apartment.py
+++ b/backend/hitas/views/apartment.py
@@ -2,7 +2,7 @@ import datetime
 import uuid
 from collections import OrderedDict
 from http import HTTPStatus
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional, Union, Iterable
 
 from django.core.exceptions import ValidationError
 from django.db.models import Prefetch, Q
@@ -34,7 +34,7 @@ from hitas.models import (
 )
 from hitas.models._base import HitasModelDecimalField
 from hitas.models.apartment import ApartmentMarketPriceImprovement, ApartmentState, DepreciationPercentage
-from hitas.utils import RoundWithPrecision, this_month
+from hitas.utils import RoundWithPrecision, this_month, valid_uuid
 from hitas.views.codes import ReadOnlyApartmentTypeSerializer
 from hitas.views.ownership import OwnershipSerializer
 from hitas.views.utils import (
@@ -44,6 +44,7 @@ from hitas.views.utils import (
     HitasModelViewSet,
     UUIDRelatedField,
     ValueOrNullField,
+    UUIDField,
 )
 from hitas.views.utils.merge import merge_model
 from hitas.views.utils.pdf import get_pdf_response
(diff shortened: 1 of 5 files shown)


backend/hitas/views/apartment.py
```diff
@@ -141,8 +142,74 @@ def validate(self, data):
...
+        apartments: Iterable[dict[str, Any]] = (
+            Apartment.objects.annotate(
+                _housing_company=Subquery(
+                    HousingCompany.objects.filter(real_estates__buildings__uuid=building_id).values("uuid"),
+                ),
+            )
+            .filter(apartment_filter)
+            .only("street_address", "apartment_number", "share_number_start", "share_number_end")
+            .values("street_address", "apartment_number", "share_number_start", "share_number_end")
+        )
```
ranta, line 166: While this is working as expected, I don't like that checking for overlapping share numbers are done in Python (requires us to save ALL of the Housing Company's apartments to memory, and loop over them), instead we could query for overlapping share numbers directly from the database.

e.g.:
```python
## Validate overlapping share numbers
# If updating, exclude the apartment being created
apartment_filter = Q()
if self.parent.instance is not None:
    apartment_filter = ~Q(uuid=self.parent.instance.uuid)

building_id = self.parent.initial_data["building"].get("id")
housing_company_uuid = (
    HousingCompany.objects.filter(real_estates__buildings__uuid=building_id).only("uuid").first().uuid
)

overlapping_apartments = (
    Apartment.objects.filter(apartment_filter)
    .filter(building__real_estate__housing_company__uuid=housing_company_uuid)
    .filter(
        Q(share_number_start__lte=start, share_number_end__gte=start)  # Overlapping start
        | Q(share_number_start__lte=end, share_number_end__gte=end)  # Overlapping end
        | Q(share_number_start__gte=start, share_number_end__lte=end)  # Fully inside range
    )
)
```

Then, when returning the overlapping apartment addresses, you can use `apartment.address`, (which also includes apartment stair)

backend/hitas/views/apartment.py
```diff
@@ -141,8 +142,74 @@ def validate(self, data):
...
             raise ValidationError({"start": "'shares.start' must not be greater than 'shares.end'."})
 
+        apartment_filter = Q(building__real_estate__housing_company__uuid=F("_housing_company"))
+        if self.parent.instance is not None:
+            apartment_filter &= ~Q(uuid=self.parent.instance.uuid)
+
+        building_id: Optional[str] = None
+        building_input = self.parent.initial_data["building"]
+        if isinstance(building_input, dict):
+            building_id = building_input.get("id")
```
ranta, line 152: Apartment will always have a `building_id` so no need to check if it exists

ranta (changes requested):
I like how all of the overlapping apartments are returned 👍 

While we don't have the capability to list all of them in the frontend (yet?), but the first one is still displayed and all of the rest can be viewed in the network tab of the browser (if it's required to see them)

backend/hitas/views/apartment.py
matti-lamppu: These checks were added here after I ran into some issues with pre-existing tests: https://github.com/City-of-Helsinki/hitas/blob/master/backend/hitas/tests/apis/test_api_housing_company.py#L565

Could just remove the test cases if these aren't real checks that need to be made, but then giving `building_id` in an invalid form would crash instead of giving a useful error. What do you think?

backend/hitas/views/apartment.py
matti-lamppu: Yes, I think we can filter out just the apartments we need from the database, I'll refactor.

One note on the current query, it only gets the "street_address", "apartment_number", "share_number_start", and "share_number_end" as a QuerySet containing a dict with `values()`, not the whole object, so the memory required is not as bad.

matti-lamppu: This change is in the latest commit

ranta (approved):
`check_share_ranges` could probably be slightly cleaned up, but we can go on with this :)

matti-lamppu:
`check_share_ranges` is still used to clamp the range of overlapping shares and add the errors for correct fields. The only thing it no longer needs is the check for if the overlap exists, but I think it can be left there just in case.

PR #193, opened 2023-01-05
Change foreign key on_delete from CASCADE to PROTECT on some models

matti-lamppu:
# Hitas Pull Request

# Description

These foreign key relationships are now protected:
- Real estate -> Housing Company
- Building -> Real Estate
- Apartments -> Building

Corresponding 409 errors have been added when trying to delete the models with references via the API.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [ ] Trying to delete a housing company with real estates now raises a 409 error
    - This can be tested with a DELETE request to  `/api/v1/housing-companies/{housing_company_id}`
- [ ] Trying to delete a real estate with buildings now raises a 409 error (the frontend currently also prevents this)
    - This can be tested with a DELETE request to  `/api/v1/housing-companies/{housing_company_id}/real_estate/{real_estate_id}`
- [ ] Trying to delete a building with apartments now raises a 409 error
    - This can be tested with a DELETE request to  `/api/v1/housing-companies/{housing_company_id}/real_estate/{real_estate_id}/buildings/{building_id}`
    - **Frontend will now silently fail on the building screen!** Should fix in another ticket.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-386


backend/hitas/migrations/0047_auto_20230105_1357.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 3.2.16 on 2023-01-05 11:57
```
ranta, line 1: Please use a verbose name the migration :)
e.g.: `./manage.py makemigrations hitas --name change_on_delete_rules`

backend/hitas/views/building.py
```diff
@@ -50,12 +51,21 @@ class BuildingViewSet(HitasModelViewSet):
...
+
     def get_queryset(self):
         hc_id = self._lookup_model_id_by_uuid(HousingCompany, "housing_company_uuid")
         re_id = self._lookup_model_id_by_uuid(RealEstate, "real_estate_uuid", housing_company_id=hc_id)
 
         return (
             Building.objects.filter(real_estate__id=re_id)
+            .prefetch_related(
+                Prefetch("apartments", Apartment.objects.only("uuid")),
+            )
```
ranta, line 68: This prefetch is adding one extra apartments db query for each apartment the building has 😬 
Why should we prefetch the Apartment uuid's anyway? 🤔

ranta (changes requested):

backend/hitas/views/building.py
matti-lamppu: Prefetch only adds [one query for all apartments](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#:~:text=Returns%20a%20QuerySet%20that%20will%20automatically%20retrieve%2C%20in%20a%20single%20batch%2C%20related%20objects%20for%20each%20of%20the%20specified%20lookups.).

The idea here was have the apartments already fetched in the wueryst, so that `instance.apartments.exists():` during DELETE wouldn't make another database query, but I think it doesn't make sense in hindsight since it's only needed for the DELETE endpoint, and the number of queries is the same either way, so I'll remove it.

backend/hitas/migrations/0047_auto_20230105_1357.py
matti-lamppu: This change is in the latest commit.

backend/hitas/views/building.py
matti-lamppu: This change is in the latest commit.

ranta (approved):
LGTM 🦚

PR #200, opened 2023-01-10
Adds onWheel event handler to number type inputs

vincit-matu:
# Hitas Pull Request

# Description

This _should_ take care of the issue where number type input value was affected by the mouse wheel.

## Pull request checklist

- **Frontend**

## Test plan

Go to any number type input and try to affect the value with the mouse wheel.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-387

### DIFF
--- a/frontend/src/common/components/formInputField/FormNumberInputField.tsx
+++ b/frontend/src/common/components/formInputField/FormNumberInputField.tsx
@@ -31,11 +31,13 @@ export default function FormNumberInputField({
     if (fractionDigits === 0) bannedCharacters.push(".");
 
     const handleWheel = (e) => {
-        e.target.blur();
-        e.stopPropagation();
-        setTimeout(() => {
-            e.target.focus();
-        }, 0);
+        if (document.activeElement === e.target) {
+            e.target.blur();
+            e.stopPropagation();
+            setTimeout(() => {
+                e.target.focus();
+            }, 0);
+        }
     };
 
     return (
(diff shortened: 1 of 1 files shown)


frontend/src/common/components/formInputField/FormNumberInputField.tsx
```diff
@@ -30,12 +30,21 @@ export default function FormNumberInputField({
     const bannedCharacters = ["e", "E", "+", "-"];
     if (fractionDigits === 0) bannedCharacters.push(".");
 
+    const handleWheel = (e) => {
+        e.target.blur();
+        e.stopPropagation();
+        setTimeout(() => {
+            e.target.focus();
+        }, 0);
+    };
```
ranta, line 39: Even with this the "input is focused/unfocused" animation is spammed on the element when scrolling very slowly, but still, this simply seems like the best way to disable scrolling on the NumberInput.

Something I would add to this however, is checking if the current element is focused before running the function (`if (document.activeElement === e.target) {...}`).
This will stop changing the focus to a different number input when scrolling (and there is no need to run the function if user is scrolling and cursor is. over an unfocused input)

ranta (approved):

frontend/src/common/components/formInputField/FormNumberInputField.tsx
vincit-matu: Good observation! Added!

PR #205, opened 2023-01-17
Sales page mockup

vincit-matu:
# Hitas Pull Request

# Description

Adds a mockup of the sales page, along with a mockup of the sales conditions display on apartment details page. Makes OwnershipsList into its own component. Lots and lots of small style fixes and sass refactoring in general.

## Pull request checklist

- **Frontend**

## Test plan

Go to an apartment detail page, view the glorious "Myyntiehdot" box. Click on "Uusi kauppa" to view the sales page mockup. Functionalitywise it should show the correct info, and allow you to make a max price calculation on the same page.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-54


frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -141,7 +104,7 @@ const LoadedApartmentSalesPage = ({apartment}: {apartment: IApartmentDetails}) =
         });
         setIsModalVisible(true);
     };
-
+    console.log(apartment, maxPriceQuery, maxPriceError);
```
ranta, line 107: Unexpected console statement :)

frontend/src/features/apartment/ApartmentSalesConditionsPage.jsx
```diff
@@ -0,0 +1,12 @@
+import {Heading, NavigateBackButton} from "../../common/components";
```
ranta, line 1: This file is accidentally `.jsx` instead of `.tsx`

frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -0,0 +1,296 @@
...
+        dateOfCount:
+            apartment.prices.maximum_prices.confirmed !== null
+                ? apartment.prices.maximum_prices.confirmed.calculation_date
+                : null,
+        price: null,
+        apartment_share_of_housing_company_loans: hasValidCalculation ? apartment.prices.primary_loan_amount : null,
+        mortgageShare: null,
+        isExcludedFromStatistics: false,
+        index: "",
+    };
```
ranta, line 59: Let's use the same names for variables across the whole application

`announcementDate` => `notificationDate` (`notification_date` in backend)
`dealDate` => `apartmentSaleDate` (Also leave this one empty if the confirmed max price calculation is not valid anymore)
`dateOfCount` This doesn't belong in `formData` as it doesn't have a field
`price` => `purchasePrice` (`purchase_price`)
`mortgageShare` this one can be removed, as its never used
`isExcludedFromStatistics` backend has this one as `include_in_statistics`, the opposite of this, so we would either need to change the naming or invert the field when sending data to the API (can be messy) or we change the naming in backend (This might be the preferred route?)
`index` not used, can be removed

frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -0,0 +1,296 @@
...
+    });
+    const [maxPrices, setMaxPrices] = useState({
+        maxPrice: apartment.prices.maximum_prices.confirmed?.maximum_price,
+        maxPricePerSquare: apartment.prices.maximum_prices.confirmed?.maximum_price
+            ? apartment.prices.maximum_prices.confirmed?.maximum_price / apartment.surface_area
+            : 0,
+        debtFreeMaxPrice: apartment.prices.debt_free_purchase_price,
+        loanAmount: apartment.prices.primary_loan_amount,
+        index: data ? data.index : maxPriceQuery?.index,
+    });
```
ranta, line 79: Same as above:

**_(suggested names are for JS naming scheme, but to simplify the process of sending data to the API, you can use the same variable names in formData as we have done in most other forms)_**

`maxPrice` => `maximumPrice`
`maxPricePerSquare` We should refrain from doing these calculations on the frontend, but this one is not yet returned from the API, created a ticket for this (HT-411)
`debtFreeMaxPrice` => `debtFreePurchasePrice`
`loanAmount` => `primaryLoanAmount`

frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -0,0 +1,296 @@
...
+    return (
+        <div className="view--apartment-sales">
+            <QueryStateHandler
+                data={data}
+                error={error}
+                isLoading={isLoading}
+            >
+                {isLoading ? (
+                    <LoadingSpinner />
+                ) : (
```
ranta, line 282: `QueryStateHandler` already handles showing the `LoadingSpinner`

frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -0,0 +1,296 @@
...
+                    />
+                </Fieldset>
+            </div>
+            <div className="row row--buttons">
+                <NavigateBackButton />
+                <SaveButton
+                    onClick={() => {
+                        return;
+                    }}
+                    isLoading={false}
```
ranta, line 229: When we pass `isLoading={isLoading}` we can display the loading indicator on the `SaveButton`'s icon. If its not required the parameter should simply be left out (The component needs to be changed so it has a default value)

frontend/src/features/apartment/ApartmentSalesPage.tsx
```diff
@@ -0,0 +1,296 @@
...
+                            label="Osuus yhtiön lainoista"
+                            unit="€"
+                            fractionDigits={2}
+                            fieldPath="apartment_share_of_housing_company_loans"
+                            formData={formData}
+                            setFormData={setFormData}
+                            error={error}
+                        />
+                    </div>
+                    <Checkbox
```
ranta, line 160: Should we instead have this?
```
<FormInputField
    inputType="checkbox"
    ...
/>
```

ranta (changes requested):

frontend/src/features/apartment/ApartmentSalesPage.tsx
vincit-matu: Possibly, but I feel this is a bit out of the scope of this **mockup**. When we do the validation I think it'll be necessary to rethink the `FormInput..`s anyway.

ranta (approved):
Requires renaming of a field in backend, but this PR is good to go :)

PR #218, opened 2023-01-31
Add sell_by_date to apartment endpoint based on its conditions of sale

matti-lamppu:
# Hitas Pull Request

# Description

Adds `sell_by_date` to apartment endpoint based on its conditions of sale.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automatic tests
- [x] Create conditions of sale for apartments with different grace periods and look at the `sell_by_date` on apartment endpoints. `sell_by_date` should always be determined based on the new apartment in the condition of sale.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-420

### DIFF
--- a/backend/hitas/views/apartment.py
+++ b/backend/hitas/views/apartment.py
@@ -1,6 +1,7 @@
 import datetime
 import uuid
 from http import HTTPStatus
+from itertools import chain
 from typing import Any, Dict, Optional, Union, Iterable
 
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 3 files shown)


backend/hitas/views/apartment.py
```diff
@@ -492,20 +493,40 @@ class ApartmentDetailSerializer(EnumSupportSerializerMixin, HitasModelSerializer
...
+    def get_sell_by_date(instance: Apartment) -> Optional[datetime.date]:
+        sell_by_dates: set[datetime.date] = set()
 
         for ownership in instance.ownerships.all():
-            conditions_of_sale_new = ownership.conditions_of_sale_new.all()
-            for cond in conditions_of_sale_new:
-                conditions_of_sale.append(cond)
+            for cos in chain(ownership.conditions_of_sale_new.all(), ownership.conditions_of_sale_old.all()):
+                if cos.fulfilled:
+                    continue
```
ranta, line 515: Shouldn't this already not include fulfilled conditions of sale?
Alternatively we can just add a `.filter(deleted__isnull=True)` to the querysets to not include them

ranta (approved):

backend/hitas/views/apartment.py
matti-lamppu: The queryset does include fulfilled conditions of sale for some time after they are fulfilled [based on a setting](https://github.com/City-of-Helsinki/hitas/blob/master/backend/hitas/models/condition_of_sale.py#L71) since this endpoint is using `condition_of_sale_queryset()`. This is useful for showing recently fulfilled conditions of sale as per the design.

Also, we shouldn't filter them here further since that will invalidate the prefetch cache (I think).

PR #220, opened 2023-02-07
Add ability to not create conditions of sale for some owners

matti-lamppu:
# Hitas Pull Request

# Description

Add ability to not create conditions of sale for some owners.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automated tests
- [x] Can set `create_conditions_of_sale` in admin panel
- [x] When creating conditions of sale for owners with the `create_conditions_of_sale=False`, no conditions of sale are created

## Tickets

This pull request resolves all or part of the following ticket(s): HT-441


backend/hitas/views/condition_of_sale.py
```diff
@@ -140,11 +141,9 @@ def get_household(owner_uuids: list[uuid.UUID]) -> list[Owner]:
                             .order_by("purchase_date")
                             .values_list("id", flat=True)[:1]
                         )
-                    ).only("id", "purchase_date"),
+                    ),
                 ),
-            )
-            .only("id")
-            .filter(uuid__in=owner_uuids)
+            ).filter(uuid__in=owner_uuids)
```
matti-lamppu, line 146: Unfortunately, I don't think we can filter out the owners with `create_conditions_of_sale=False` here, since we have a check after this whether we got all the owners we were supposed to. There is no way to know whether the owner was not fetched because the owner id was wrong, or if it should not have conditions of sale created for it.

backend/hitas/tests/factories/owner.py
```diff
@@ -45,6 +45,7 @@ class Meta:
     name = factory.Faker("name")
     identifier = faker.identifier()
     email = factory.Faker("email")
+    create_conditions_of_sale = True
```
ranta, line 48: `create_conditions_of_sale` sounds like a method 😅 
I would have gone with maybe something such as `bypass_conditions_of_sale`

ranta (approved):

backend/hitas/tests/factories/owner.py
matti-lamppu: I'll change the name, it might get confusing.

PR #222, opened 2023-02-07
Fix latest sales migration

ranta:
# Hitas Pull Request

# Description

- Add a missing `not` to the conditional, which was incorrectly deleting new ownerships and keeping the old ones.
- Order Housing Companies in the API by completed on date instead of ID. This way newest companies are listed first by default which is the preferred behaviour

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Run the migration on your local pc to find that no duplicate ownerships are created anymore :)

## Tickets

This pull request resolves all or part of the following ticket(s): -


matti-lamppu (approved):
👍

PR #224, opened 2023-02-09
Create conditions of sale on apartment sale

matti-lamppu:
# Hitas Pull Request

# Description

Now making a sale for a new apartment for an owner with previous ownerships will automatically create conditions of sale between the owner's apartments.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automatic tests
- [x] Create a sale for a new apartment with an owner that also owns other apartments. Conditions of sale should be created between those apartments with the sale, and a flag set in the sale response to indicate this.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-74

### DIFF
--- a/backend/hitas/views/apartment_sale.py
+++ b/backend/hitas/views/apartment_sale.py
@@ -87,13 +87,22 @@ class ApartmentSaleCreateSerializer(HitasModelSerializer):
                     Ownership.objects.select_related("apartment"),
                 ),
                 # Limit the fetched sales to only the first sale,
-                # as we only need that to figure out if the apartment is new
-                prefetch_first_sale("ownerships__apartment__"),
+                # as we only need that to figure out if the apartment is new.
+                # Ignore the sale we just created so that this apartment is treated as new.
+                prefetch_first_sale("ownerships__apartment__", ignore=[instance.id]),
             )
             for owner in owners:
                 cos = create_conditions_of_sale(owners=[owner])
                 if cos:
                     self.context["conditions_of_sale_created"] = True
+
+            apartment.first_purchase_date = instance.purchase_date
+            apartment.save()
+            return instance
+
+        if apartment.latest_purchase_date is None or instance.purchase_date > apartment.latest_purchase_date:
+            apartment.latest_purchase_date = instance.purchase_date
+            apartment.save()
 
         return instance
 
--- a/backend/hitas/views/apartment_sale.py
+++ b/backend/hitas/views/apartment_sale.py
@@ -1,11 +1,15 @@
 from typing import Any
 
 from django.core.exceptions import ValidationError
-from django.db.models import Prefetch
+from django.db.models import Prefetch, prefetch_related_objects
+from rest_framework import serializers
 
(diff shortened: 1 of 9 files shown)


backend/hitas/views/apartment_sale.py
```diff
@@ -87,14 +87,23 @@ def create(self, validated_data: dict[str, Any]) -> ApartmentSale:
                     Ownership.objects.select_related("apartment"),
                 ),
                 # Limit the fetched sales to only the first sale,
-                # as we only need that to figure out if the apartment is new
-                prefetch_first_sale("ownerships__apartment__"),
+                # as we only need that to figure out if the apartment is new.
+                # Ignore the sale we just created so that this apartment is treated as new.
+                prefetch_first_sale("ownerships__apartment__", ignore=[instance.id]),
```
ranta, line 92: I'm having a really hard time understanding why we need to consider the apartment as new when it has been sold already

backend/hitas/views/apartment_sale.py
```diff
@@ -87,14 +87,23 @@ def create(self, validated_data: dict[str, Any]) -> ApartmentSale:
...
                 if cos:
                     self.context["conditions_of_sale_created"] = True
 
+            apartment.first_purchase_date = instance.purchase_date
+            apartment.save()
+            return instance
+
+        if apartment.latest_purchase_date is None or instance.purchase_date > apartment.latest_purchase_date:
+            apartment.latest_purchase_date = instance.purchase_date
+            apartment.save()
```
ranta, line 105: Is this last commit about something to do with _conditions of sale_ like the name suggests, or is it about setting the purchase dates? 🤔

backend/hitas/views/apartment_sale.py
matti-lamppu: It's because of this edge case:
1. Apartment has been completed AND has not been sold before -> it's still considered NEW
2. We made the first sale above AND **_it is in the past_** -> Rules now count the apartment as NOT NEW
3. If we would try to create conditions of sale for the new ownerships, this apartment that was just sold would not be identified as NEW, and no conditions of sale would be created for it.

Ideally `apartment.is_new` would be able to recognize apartments like this as new, but they would have to be separated from apartments that have only sold once but are not new anymore. I don't think we have a clear rule for this, unless I've missed something?

backend/hitas/views/apartment_sale.py
matti-lamppu: I simply forgot that these fields have to be set (for now), so I put them in the commit. I guess I could've separated them into their own commit.

backend/hitas/views/apartment_sale.py
matti-lamppu: Note that this only starts to matter once we remove `first_purchase_date` from apartment. Currently, I have `first_purchase_date` be set after the conditions of sale are created, which works out so that the apartment is new regardless of whether we ignore the previously created sale, since both the `first_purchase_date` and the first sale are checked in `apartment.is_new`.

ranta (approved):

PR #226, opened 2023-02-09
Add sale by date to apartment list endpoint

matti-lamppu:
# Hitas Pull Request

# Description

Adds two fields: `sell_by_date` and `on_grace_period` to the apartment list/search endpoint.
- Without conditions of sale on any of the apartments, the value for `sell_by_date` should be null, and `on_grace_period` False
- With conditions of sale on any of the apartments, the value for `sell_by_date` will be the earliest of them (taking the grace period into account), and `on_grace_period` will be True if any of the conditions of sale for an apartment have one set, and false otherwise.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automated tests

## Tickets

This pull request resolves all or part of the following ticket(s): HT-443

### DIFF
--- a/backend/hitas/models/apartment.py
+++ b/backend/hitas/models/apartment.py
@@ -133,6 +133,7 @@ class Apartment(ExternalHitasModel):
 
         return no_first_purchase_or_in_the_future and no_first_sale_or_in_the_future
 
+    @property
     def sell_by_date(self):
         sell_by_dates: set[datetime.date] = set()
 
@@ -157,7 +158,8 @@ class Apartment(ExternalHitasModel):
 
         return min(sell_by_dates)
 
-    def on_grace_period(self) -> bool:
+    @property
+    def has_grace_period(self) -> bool:
         for ownership in self.ownerships.all():
             for cos in chain(ownership.conditions_of_sale_new.all(), ownership.conditions_of_sale_old.all()):
                 if cos.fulfilled:
--- a/backend/hitas/models/apartment.py
+++ b/backend/hitas/models/apartment.py
@@ -157,6 +157,15 @@ class Apartment(ExternalHitasModel):
 
         return min(sell_by_dates)
 
+    def on_grace_period(self) -> bool:
+        for ownership in self.ownerships.all():
+            for cos in chain(ownership.conditions_of_sale_new.all(), ownership.conditions_of_sale_old.all()):
+                if cos.fulfilled:
+                    continue
+                if cos.grace_period != GracePeriod.NOT_GIVEN:
+                    return True
+        return False
+
     class Meta:
         verbose_name = _("Apartment")
         verbose_name_plural = _("Apartments")
--- a/backend/hitas/models/apartment.py
+++ b/backend/hitas/models/apartment.py
@@ -1,7 +1,10 @@
+import datetime
 import uuid
 from decimal import Decimal
+from itertools import chain
 from typing import Optional
 
+from dateutil.relativedelta import relativedelta
 from django.core.validators import MinValueValidator
 from django.db import models
 from django.utils import timezone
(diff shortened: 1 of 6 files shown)


backend/hitas/tests/apis/test_api_apartment_list.py
```diff
@@ -144,6 +153,157 @@ def test__api__apartment__list(api_client: HitasAPIClient):
...
+                    "owner": {
+                        "id": o3.owner.uuid.hex,
+                        "name": o3.owner.name,
+                        "identifier": o3.owner.identifier,
+                        "email": o3.owner.email,
+                    },
+                    "percentage": float(o3.percentage),
+                }
+            ],
+            "sell_by_date": str(ap2.completion_date),
```
ranta: In this test we are only actually interested in the `sell_by_date` value, so this json could be shortened to only test for that :)
(Would make this a little bit more readable, and require less maintenance when we change the API)

backend/hitas/models/apartment.py
```diff
@@ -129,6 +133,30 @@ def is_new(self) -> bool:
 
         return no_first_purchase_or_in_the_future and no_first_sale_or_in_the_future
 
+    def sell_by_date(self):
```
ranta, line 136: This should be changed to be a `property` or name be should changed to `get_sell_by_date` to signify that this is actually a method

backend/hitas/models/apartment.py
```diff
@@ -129,6 +133,30 @@ def is_new(self) -> bool:
...
+        sell_by_dates: set[datetime.date] = set()
+
+        for ownership in self.ownerships.all():
+            for cos in chain(ownership.conditions_of_sale_new.all(), ownership.conditions_of_sale_old.all()):
```
ranta, line 140: We are starting to have a bit of these related model accessions, that assume the related objects are cached.
In the future might want to consider creating tests for these, to make sure they are actually already fetched, and dont cause unnecessary db queries. What do you think? 🤔

backend/openapi.yaml
```diff
@@ -3921,6 +3922,9 @@ components:
           format: date
           nullable: true
           readOnly: true
+        on_grace_period:
+          description: Do any of the conditions of sale related to this apartment have a grace period set?
+          type: boolean
```
ranta, line 3927: If we wanted to describe if this apartment is currently on grace period, the field should be renamed to such that it hint that outputs a boolean value: `is_on_grace_period`.

But now this field is a bit confusingly named i think,  as it doesn't tell if the apartment is currently on grace period, it tells does it have any grace period currently given, and so we should rename it to e.g. `has_grace_period_set` or something that does describe what it actually is.

ranta (changes requested):

backend/openapi.yaml
matti-lamppu: I can see how it might be confusing, changed it to just `has_grace_period`.

backend/hitas/models/apartment.py
matti-lamppu: I made a context manager test tool that can be used to count the database queries happening in its context. Here is an example of it's usage:

```python
# Expect that calling this endpoint will make 5 database queries
with count_queries(5):
    response = api_client.get(reverse("hitas:apartment-list"))
```

Let me know what you think. There is also the tedious part of going through every test and adding this context manager to them...

backend/hitas/models/apartment.py
matti-lamppu: Made `sell_by_date` and `has_grace_period` into properties.

backend/hitas/tests/apis/test_api_apartment_list.py
matti-lamppu: Simplified the test.

backend/hitas/models/apartment.py
ranta: I didn't think this would have been a part of this ticket, but ok sure 😆 

> There is also the tedious part of going through every test and adding this context manager to them
And also figuring out what is the correct amount of queries for the context 🙂

ranta (approved):
🌟

PR #229, opened 2023-02-16
Sort apartments with conditions of sale first in apartment endpoints

matti-lamppu:
# Hitas Pull Request

# Description

Apartments are now sorted based on if they have conditions of sale to their ownerships, and those that have some come first.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automatic tests
- [x] Add conditions of sale to some apartments, look at the apartments lists and observe the ordering

## Tickets

This pull request resolves all or part of the following ticket(s): HT-455

### DIFF
--- a/backend/hitas/tests/apis/test_api_apartment.py
+++ b/backend/hitas/tests/apis/test_api_apartment.py
@@ -227,6 +227,34 @@ def test__api__apartment__list(api_client: HitasAPIClient):
 
 
 @pytest.mark.django_db
+def test__api__apartment__list__those_with_conditions_of_sale_come_first(api_client: HitasAPIClient):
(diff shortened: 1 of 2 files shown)


ranta (changes requested):
We should order the apartments also based on the `sell_by_date`

matti-lamppu:
`sell_by_date` is calculated after the queryset is evaluated, so I would need to refactor it somehow to an annotation, which would be pretty hard given the complexity of how `sell_by_date` is determined. Can't use `sorted()` either, since pagination expects a queryset. Asking if it is really necessary :D

ranta (approved):
There will never be that many apartments with conditions of sale anyway, so yes we can live without it.
We can always spend more time on this at a later time if it is really necessary 😄

PR #230, opened 2023-02-16
Fix edge case for new apartment sell by dates

matti-lamppu:
# Hitas Pull Request

# Description

Fixes `sell_by_date` calculation in cases where the apartment is completed before it is sold for the first time. 
In these cases, the `sell_by_date` should be calculated based on the first sale purchase date.

A bit of refactoring is also made to reuse code and avoid import loops.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automatic tests

## Tickets

This pull request resolves all or part of the following ticket(s): HT-459

### DIFF
--- a/backend/hitas/services/condition_of_sale.py
+++ b/backend/hitas/services/condition_of_sale.py
@@ -1,0 +1,122 @@
+from django.conf import settings
+from django.db import models
+from django.db.models import Q
+from django.utils import timezone
+
+from hitas.models import ConditionOfSale, Owner, Ownership
+from hitas.models.condition_of_sale import ConditionOfSaleAnnotated
+from hitas.services.apartment import subquery_first_purchase_date
+
+
+def condition_of_sale_queryset() -> models.QuerySet[ConditionOfSaleAnnotated]:
+    return (
+        ConditionOfSale.all_objects.filter(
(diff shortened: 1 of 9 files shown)


backend/hitas/services/condition_of_sale.py
```diff
@@ -0,0 +1,122 @@
...
+from django.db import models
+from django.db.models import Q
+from django.utils import timezone
+
+from hitas.models import ConditionOfSale, Owner, Ownership
+from hitas.models.condition_of_sale import ConditionOfSaleAnnotated
+from hitas.services.apartment import subquery_first_purchase_date
+
+
+def condition_of_sale_queryset() -> models.QuerySet[ConditionOfSaleAnnotated]:
```
ranta, line 11: Refactoring code should be in their own commits, not mixed in the same commit with unrelated changes.

ranta (approved):

backend/hitas/services/condition_of_sale.py
matti-lamppu: I refactored these because I initially had the apartment and condition of sale functions in the respective model's files, but that caused import loops, so I had to move both of them to a new place.

PR #231, opened 2023-02-17
Conditions of sale

ranta:
# Hitas Pull Request

# Description

- Conditions of Sale listing page
- Manually creating conditions of sale
- Conditions of sale filter on apartment listing page
- Refactor apartment listing page backend
- Bug fixes and improvements

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- Manually create conditions of sale
- Check that description items seem to be working correctly

## Tickets

This pull request resolves all or part of the following ticket(s): HT-170, HT-421, HT-174

### DIFF
--- a/backend/hitas/calculations/max_prices/rules_2011_onwards.py
+++ b/backend/hitas/calculations/max_prices/rules_2011_onwards.py
@@ -14,14 +14,16 @@ from hitas.models.apartment import ApartmentWithAnnotationsMaxPrice
 
 class Rules2011Onwards(CalculatorRules):
     def validate_indices(self, apartment: ApartmentWithAnnotationsMaxPrice) -> None:
-        if (
-            apartment.calculation_date_cpi_2005eq100 is None
-            or apartment.completion_date_cpi_2005eq100 is None
-            or apartment.calculation_date_mpi_2005eq100 is None
-            or apartment.completion_date_mpi_2005eq100 is None
-            or apartment.surface_area_price_ceiling is None
-        ):
-            raise IndexMissingException()
+        if apartment.calculation_date_cpi_2005eq100 is None:
+            raise IndexMissingException("calculation_date_cpi_2005eq100")
+        if apartment.completion_date_cpi_2005eq100 is None:
+            raise IndexMissingException("completion_date_cpi_2005eq100")
+        if apartment.calculation_date_mpi_2005eq100 is None:
+            raise IndexMissingException("calculation_date_mpi_2005eq100")
+        if apartment.completion_date_mpi_2005eq100 is None:
+            raise IndexMissingException("completion_date_mpi_2005eq100")
+        if apartment.surface_area_price_ceiling is None:
+            raise IndexMissingException("surface_area_price_ceiling")
 
     def calculate_construction_price_index_max_price(
         self,
@@ -107,7 +109,7 @@ class Rules2011Onwards(CalculatorRules):
         max_price = debt_free_shares_price - apartment_share_of_housing_company_loans
 
         if max_price <= 0:
-            raise InvalidCalculationResultException()
+            raise InvalidCalculationResultException("max_price_lte_zero")
 
         return IndexCalculation(
             maximum_price=max_price,
--- a/backend/hitas/calculations/max_prices/rules_2011_onwards.py
+++ b/backend/hitas/calculations/max_prices/rules_2011_onwards.py
@@ -13,17 +13,17 @@ from hitas.models.apartment import ApartmentWithAnnotationsMaxPrice
 
 
 class Rules2011Onwards(CalculatorRules):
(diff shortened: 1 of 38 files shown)


backend/hitas/calculations/max_prices/rules_2011_onwards.py
```diff
@@ -14,14 +14,16 @@
...
     def validate_indices(self, apartment: ApartmentWithAnnotationsMaxPrice) -> None:
-        if (
-            apartment.calculation_date_cpi_2005eq100 is None
-            or apartment.completion_date_cpi_2005eq100 is None
-            or apartment.calculation_date_mpi_2005eq100 is None
-            or apartment.completion_date_mpi_2005eq100 is None
-            or apartment.surface_area_price_ceiling is None
-        ):
-            raise IndexMissingException()
+        if apartment.calculation_date_cpi_2005eq100 is None:
```
matti-lamppu, line 17: 👍👍

backend/hitas/views/apartment.py
```diff
@@ -676,11 +703,21 @@ class Meta:
...
+    has_conditions_of_sale = serializers.SerializerMethodField()
+    has_grace_period = serializers.SerializerMethodField()
 
     @staticmethod
     def get_type(instance: Apartment) -> Optional[str]:
         return getattr(getattr(instance, "apartment_type", None), "value", None)
 
+    @staticmethod
+    def get_has_conditions_of_sale(instance: Apartment) -> bool:
+        return instance.has_conditions_of_sale  # Annotated field
```
matti-lamppu, line 715: You can annotate this method with `hitas.models.apartment.ApartmentWithAnnotations` and add the `has_conditions_of_sale` there.

backend/hitas/views/apartment_list.py
```diff
@@ -1,149 +1,18 @@
...
-            .annotate(
-                sales_condition=Case(
-                    When(condition=Q(number_of_conditions_of_sale__gt=0), then=True),
-                    default=False,
-                    output_field=models.BooleanField(),
-                ),
-            )
-            .order_by("apartment_number", "id")
-        )
+        return ApartmentViewSet.get_base_queryset()
```
matti-lamppu, line 15: Just right

matti-lamppu (approved):
@vincit-matu should also review the frontend changes

backend/hitas/calculations/max_prices/rules_pre_2011.py
```diff
@@ -42,6 +44,11 @@ def calculate_construction_price_index_max_price(
         housing_company_improvements: List,
         calculation_date: datetime.date,
     ) -> IndexCalculation:
+        if not apartment.completion_date_realized_housing_company_acquisition_price:
```
matti-lamppu, line 47: Should also check `realized_housing_company_acquisition_price` is not zero/null since `calculate_housing_company_improvements_pre_2011_construction_price_index` cannot be performed without it (leads to 0 division error).

backend/hitas/views/apartment_max_price.py
```diff
@@ -42,21 +42,21 @@ def create(self, request, *args, **kwargs):
                 additional_info=serializer.validated_data.get("additional_info", ""),
             )
             return Response(max_prices)
-        except InvalidCalculationResultException:
+        except InvalidCalculationResultException as e:
             return Response(
```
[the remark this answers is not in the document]
ranta: Changed the exceptions to inherit from `ModelConflict` and added even more verbosity to the error messages (now they now return the date) 😄

frontend/src/features/apartment/ApartmentCreatePage.tsx
```diff
@@ -179,7 +165,7 @@ const ApartmentCreatePage = () => {
         }
     };
 
-    const [formData, setFormData] = useImmer<IApartmentWritable>(initialFormData);
+    const [formData, setFormData] = useImmer<IApartmentWritableForm>(initialFormData);
```
vincit-matu, line 168: Hmm.. I think there's no need for immer now that we're using react-hook-form. It state-variable itself seems to be used here only to load the initial value, so using `initialFormData` as is would probably be enough.
And the setter-function is only called upon a successful save operation.  When the form is submitted successfully it updates the values in the API automagically, so we have a new `initialFormData` anyway. I think there's no need to have a state-variable take care of the same thing.

vincit-matu (changes requested):
Looks fine, apart from using the immer-liNrary when a component otherwise makes use of the react-hook-form.  Tje rstem. Looks good to me!

frontend/src/features/apartment/ApartmentCreatePage.tsx
ranta: Good catch! I had plans to remove it, but looks like got distracted halfway through the process 😅

vincit-matu (approved):
👍

PR #234, opened 2023-02-20
Condition of Sale UI/styles

vincit-matu:
# Hitas Pull Request

# Description

Restyles the conditions of sale in the UI. Implements the HDS Pagination component, and fixes the problem with pressing tab and/or shift+tab when focused on a related model field. Before it was impossible to move the focus away from a related model input, as any keypress simply opened the relative model dialogbox.

## Pull request checklist

- **Frontend**

## Test plan

Check out the sales conditions in both apartment detail view, and their own page. The remove-button was moved to it's correct place and I generally also compacted the list. The only thing I couldn't test was how the possible 3/6 month grace period is shown in the list. I replaced the raw values with icons from HDS (cross for no grace period, and a check mark along with the duration for the other two possible options there. 

## Tickets

This pull request resolves all or part of the following ticket(s): HT-392

### DIFF
--- a/frontend/src/common/components/ListPageNumbers.tsx
+++ b/frontend/src/common/components/ListPageNumbers.tsx
@@ -1,44 +1,36 @@
 import React from "react";
 
-import {Button} from "hds-react";
+import {Pagination} from "hds-react";
 
 import {PageInfo} from "../schemas";
 
 interface ListPageNumbersProps {
-    pageInfo: PageInfo | undefined;
+    pageInfo: PageInfo;
     currentPage: number;
     setCurrentPage: React.Dispatch<React.SetStateAction<number>>;
 }
 
-export default function ListPageNumbers({currentPage, setCurrentPage, pageInfo}: ListPageNumbersProps): JSX.Element {
-    if (pageInfo === undefined) return <></>;
-
-    const pageButtonData: {pageNumber: number; buttonText: string | number}[] = [];
-    const totalPages = pageInfo.total_pages;
-    for (let i = -3; i <= 3; i++) {
-        if (currentPage + i > 0 && currentPage + i <= totalPages) {
-            pageButtonData.push({pageNumber: currentPage + i, buttonText: currentPage + i});
-        }
-    }
-    pageButtonData.unshift({pageNumber: Math.max(1, currentPage - 1), buttonText: "<"});
-    pageButtonData.push({pageNumber: Math.min(totalPages, currentPage + 1), buttonText: ">"});
-    if (totalPages === 1) return <></>;
+export default function ListPageNumbers({
+    currentPage,
+    setCurrentPage,
+    pageInfo,
+    pageInfo: {total_pages: totalPages},
+}: ListPageNumbersProps): JSX.Element {
+    if (pageInfo === undefined || totalPages === 1) return <></>;
     return (
-        <div style={{display: "flex", justifyContent: "space-between"}}>
-            {pageButtonData.map(({pageNumber, buttonText}) => (
-                <Button
-                    key={`pageButton__${buttonText}`}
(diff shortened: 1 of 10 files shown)


frontend/src/features/housingCompany/HousingCompanyListPage.tsx
```diff
@@ -69,11 +69,15 @@ const HousingCompanyResultsList = ({filterParams}): JSX.Element => {
...
-                />
+                {data !== undefined ? (
+                    <ListPageNumbers
+                        currentPage={currentPage}
+                        setCurrentPage={setCurrentPage}
+                        pageInfo={data.page}
+                    />
+                ) : (
+                    <></>
+                )}
```
ranta, line 80: No need to check if data is undefined, QueryStateHandler already does that, you can do this instead:
```ts
<ListPageNumbers
    currentPage={currentPage}
    setCurrentPage={setCurrentPage}
    pageInfo={(data as IHousingCompanyListResponse).page}
/>
```

frontend/src/common/components/ListPageNumbers.tsx
```diff
@@ -1,44 +1,36 @@
...
-    }
-    pageButtonData.unshift({pageNumber: Math.max(1, currentPage - 1), buttonText: "<"});
-    pageButtonData.push({pageNumber: Math.min(totalPages, currentPage + 1), buttonText: ">"});
-    if (totalPages === 1) return <></>;
+export default function ListPageNumbers({
+    currentPage,
+    setCurrentPage,
+    pageInfo,
+    pageInfo: {total_pages: totalPages},
+}: ListPageNumbersProps): JSX.Element {
```
ranta, line 18: Didn't even know this style of typing could be used 😮 
But isn't `pageInfo` typed here twice, why not just use the interface instead of retyping it like this? 🤔

frontend/src/styles/components/_ApartmentDetails.sass
```diff
@@ -42,7 +42,7 @@
         content: "|"
 
     &--number
-      padding: $spacing-xs $spacing-s
+      padding: $spacing-xs $spacing-s                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               3
```
ranta, line 45: What's happening here? 😄

ranta (changes requested):

frontend/src/features/housingCompany/HousingCompanyListPage.tsx
vincit-matu: Nice! I have to admit it didn't feel right using a ternary operator simply to please type checking and/or the linter. 😅

frontend/src/styles/components/_ApartmentDetails.sass
vincit-matu: No idea. :D I've stumbled on "changes" like this sometimes before with git.

frontend/src/common/components/ListPageNumbers.tsx
vincit-matu: Yeah, nested abstraction. 
But no, `pageInfo` is not defined here twice. It's defined on line 16, and on line 17 we only define `totalPages`. There `pageInfo` is only given as pointer to access the `total_pages` value within it, which is then abstracted to `totalPages`. It's just a handy way to abstract the value, and as I think the ListPageNumbers was using the `totalPages` variable. I think using it within the element is better/clearer than having to refer to it e.g. as `pageInfo.total_pages`. Of course it could also be declared as a constant separately, i.e. `const totalPages = pageInfo.total_pages;` - but the abstraction is more elegant imho! 🎩

ranta (approved):
LGTM 🎸

PR #236, opened 2023-02-21
Read statistics center sales from excel

matti-lamppu:
# Hitas Pull Request

# Description

Adds the ability to read tilastokeskus excel sheet to external sales data for the use in surface area price ceiling calculation, as well as the ability to fetch the data based on a quarter.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

- [x] Automatic tests
- [x] Take the example Excel-file (`Helsingin aritmeettiset neliöhinnat postialueittain 2021Q2_2022Q4-alkuperäinen.xlsx`) from HT-444 ticket in Jira, and open Swagger Editor (`localhost:8090` when docker-container is running). Add a new configuration for the development API like below so that Swagger can send data to it. Send the Excel file through the `/api/v1/external-sales-data` endpoint and observe the results. The database should save the results as well in the `hitas_externalsalesdata` table.
- [x] After using the above method to save from an Excel file, the read endpoint can be used to read based on the last quarter in the excel file (`2022Q4` in this case)

```yaml
servers:
  - url: http://localhost:<my_port>
```

## Tickets

This pull request resolves all or part of the following ticket(s): HT-444


backend/hitas/services/validation.py
```diff
@@ -0,0 +1,17 @@
...
+
+def validate_postal_code(postal_code: str) -> str:
+    match = re.match(r"^\d{5}$", postal_code)
+    if match is None:
+        raise ValidationError(f"{postal_code!r} is not a valid postal code.")
+    return postal_code
+
+
+def validate_quarter(value: str) -> str:
+    match = re.match(r"^\d{4}Q[1234]$", value)
```
ranta, line 14: These are pretty simple regexes, but in general adding an example value still helps 😉

backend/hitas/services/validation.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+from rest_framework.exceptions import ValidationError
+
+
+def validate_postal_code(postal_code: str) -> str:
```
ranta, line 6: We have already a bunch of `validate_` methods in `hitas/models/utils.py`, maybe those can be moved here as well (as a part of the general architecture improvements)

ranta (approved):
LGTM 📈 

We should check from the customer if we should include all postal codes that have a housing company, or only those that have a housing company which is still monitored in Hitas

backend/hitas/services/validation.py
matti-lamppu: Was planning to, although when I tried this some time ago there were some import problems, but I think those can be solved with some more effort.

backend/hitas/services/validation.py
matti-lamppu: Added Docstrings with example values.

matti-lamppu:
Ok, I think we can refine the saved values also later (e.g. also only those postal codes which would fall under the 30-year comparison at the moment) when we do the actual 30-year comparison feature, so I'll merge this for now.

PR #237, opened 2023-02-21
Add maximum price per square meter

vincit-matu:
# Hitas Pull Request

# Description

Adds a new value to the maximum price calculation endpoint, and refactors the front to use said value (maximum_price_per_square), instead of attempting to calculate it on its own. 

Also fixes the type checking bug which is currently breaking the frontend (== adds a missing `?` to an object reference, thus allowing for the object to be undefined on the first registering of the component). It was crashing on load since it's trying to refer to a nested attribute of an object which can not have been loaded on the initial run, but now accepts that the object in question can be undefined.

## Pull request checklist

- **Frontend**
- **Backend**

## Test plan

A) Notice that the front works instead of crashing on any/all pages (this was due to the missing `?`)
B) Go to the apartment sale page for any apartment and make a maximum price calculation. There's no drastic changes, but now the value for the maximum price per square meter comes straight from the backend. 

## Tickets

This pull request resolves all or part of the following ticket(s): HT-411, HT-392


ranta (approved):
LGTM 🔳

PR #244, opened 2023-03-13
Separate out catalog_acquisition_price in the apartment endpoint

matti-lamppu:
# Hitas Pull Request

# Description

Separate catalog prices from acquisition prices in the apartment endpoint.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**
  - [x] Terminology page in Confluence has been updated

## Test plan

- [x] Automated tests
- [x] Apartment page should still work normally, but catalog prices won't show up in the acquisition prices anymore. HT-495 will fix this behavior.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-494


backend/hitas/models/apartment.py
```diff
@@ -78,17 +78,35 @@ class Apartment(ExternalHitasModel):
...
+    def current_acquisition_price(self) -> Optional[Decimal]:
+        first_sale_price = self.first_sale_acquisition_price
+        if first_sale_price:
+            return first_sale_price
+
+        catalog_price = self.catalog_acquisition_price
+        if catalog_price:
+            return catalog_price
 
+        return None
```
ranta, line 92: This property is not used anywhere anymore (except a test). Reading the ticket's description this property seems not needed anymore, so this can be removed
To separate the catalogue/first sale values we can just do a little bit of logic in the frontend, since we already have both values there.

ranta (changes requested):

backend/hitas/models/apartment.py
matti-lamppu: Done

ranta (approved):
🦘

PR #265, opened 2023-03-30
Housing company state refactor

matti-lamppu:
# Hitas Pull Request

# Description

Replaces Housing company state with four fields: `regulation_status`, `exclude_from_statistics`, `over_thirty_years_old`, and `completed`. The first two are editable fields, while the last two are computable fields. Housing company still retains the state field until frontend changes are made.

## Pull request checklist

- **Testing**
- **Database**
  - [x] Database migrations will work in the DEV & TEST environments
- **Documentation**

## Test plan

- [x] Automated tests
- [x] Run migration without issues
- [x] All old Housing Company `state`-related stuff works with the new fields

## Tickets

This pull request resolves all or part of the following ticket(s): HT-510


backend/hitas/models/housing_company.py
```diff
@@ -95,6 +114,29 @@ def total_shares_count(self) -> int:
...
+        from hitas.models import Apartment
+
+        completion_dates: list[datetime.date | None] = Apartment.objects.filter(
+            building__real_estate__housing_company=self
+        ).values_list("completion_date", flat=True)
+
+        if None in completion_dates:
+            return False
+
+        return relativedelta(timezone.now().date(), max(completion_dates)).years >= 30
```
ranta, line 128: I would change the property name to signal the the return value is a boolean. Also, there shouldn't be a need to do the checking in python, as we can just ask the DB to return the needed value like so:

```py
    @property
    def is_over_thirty_years_old(self) -> bool:
        from hitas.models import Apartment

        completion_date: Optional[datetime.date] = (
            Apartment.objects.filter(building__real_estate__housing_company=self)
            .order_by("-completion_date")
            .first()
            .completion_date
        )

        if completion_date is None:
            return False

        return relativedelta(timezone.now().date(), completion_date).years >= 30
```

backend/hitas/models/housing_company.py
```diff
@@ -95,6 +114,29 @@ def total_shares_count(self) -> int:
...
+
+    @property
+    def completed(self) -> bool:
+        from hitas.models import Apartment
+
+        completion_dates: list[datetime.date | None] = Apartment.objects.filter(
+            building__real_estate__housing_company=self
+        ).values_list("completion_date", flat=True)
+
+        return None not in completion_dates
```
ranta, line 138: Here the same thing as above, we don't need to do the checking in python, we can simply:
```py
def is_completed(self) -> bool:
    completion_date = ...
    return bool(completion_date)
```

backend/hitas/tests/factories/housing_company.py
```diff
@@ -22,6 +23,8 @@ class Meta:
     display_name = factory.Sequence(lambda n: f"Test Housing company {n:03}")
     official_name = factory.LazyAttribute(lambda self: f"As Oy {self.display_name}")
     state = fuzzy.FuzzyChoice(state[0] for state in HousingCompanyState.choices())
+    exclude_from_statistics = False
+    regulation_status = fuzzy.FuzzyChoice(state[0] for state in RegulationStatus.choices())
```
ranta, line 27: I'd rather have the `regulation_status` default value be `RegulationStatus.REGULATED`, so it doesn't need to be specified for every case separately.
Then when we need a non-regulated housing company we can specify it.

backend/hitas/oracle_migration/utils.py
```diff
@@ -80,6 +81,18 @@ def housing_company_state_from(code: str) -> HousingCompanyState:
...
+def housing_company_regulation_status_from(code: str) -> RegulationStatus:
+    # These are hardcoded as the code number (C_KOODISTOID) and
+    # the name (C_NAME) are the only ones that can be used to
+    # identify these types
+    if code == "003":
+        return RegulationStatus.RELEASED_BY_HITAS
+    if code == "004":
+        return RegulationStatus.RELEASED_BY_PLOT_DEPARTMENT
+
+    return RegulationStatus.REGULATED
```
ranta, line 93: Should `005` (`Half-Hitas`) be regulated or not? I'm not 100% sure which one it should be 🤔

backend/hitas/views/housing_company.py
```diff
@@ -45,6 +49,18 @@
 from hitas.views.utils.serializers import YearMonthSerializer
 
 
+class RegulationStatusViewSet(ViewSet):
+    def list(self, request: Request, *args, **kwargs) -> Response:
+        data = [
+            {
+                "name": regulation_status.value,
```
ranta, line 56: `value` or `name`?

`name` sounds kind of synonymous with `label` to me 🤔

backend/hitas/oracle_migration/utils.py
matti-lamppu: If I remember correctly, half-Hitas housing companies are only regulated for the initial sale, and not afterward.

backend/hitas/tests/factories/housing_company.py
matti-lamppu: Sure, I'll change it

backend/hitas/models/housing_company.py
matti-lamppu: Yes, I'll change this.

backend/hitas/models/housing_company.py
matti-lamppu: Changed.

backend/hitas/views/housing_company.py
matti-lamppu: I guess we can go with the value-label separation enums use. I'll change it.

ranta (approved):
🐴

PR #271, opened 2023-04-03
Obfuscate owners without regulated apartments after 30-year regulation

matti-lamppu:
# Hitas Pull Request

# Description

Adds the ability to obfuscate owner information when the owner no longer owns any regulated apartments after a 30-year regulation check. The obfuscation simply sets their name, identifier, and email to null. If any owners were obfuscated this way, they are returned in the 30-year regulation check response (in their pre-obfuscation state) for the frontend to show.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] This is a bit tedious to test manually since the setup is so specific, but you can set up some scenarios in the automated tests and go through them.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-520

### DIFF
--- a/backend/hitas/services/owner.py
+++ b/backend/hitas/services/owner.py
@@ -1,0 +1,48 @@
+from django.db import models
+from django.db.models import Count, OuterRef, QuerySet
+from django.db.models.functions import Coalesce
+
+from hitas.models.housing_company import RegulationStatus
+from hitas.models.owner import Owner, OwnerT
+from hitas.models.ownership import Ownership
+from hitas.utils import SQSum
+
+
+def obfuscate_owners_without_regulated_apartments() -> list[OwnerT]:
+    """Remove any personal information from owners which do not own any regulated hitas apartments."""
+    owners: QuerySet[Owner] = Owner.objects.annotate(
+        owned_regulated_housing_companies=Coalesce(
+            SQSum(
+                queryset=(
+                    Ownership.objects.select_related(
+                        "owner",
+                        "sale__apartment__building__real_estate__housing_company",
+                    )
+                    .filter(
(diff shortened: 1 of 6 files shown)


backend/hitas/services/owner.py
```diff
@@ -0,0 +1,64 @@
...
+from django.db.models import Count, OuterRef
+from django.db.models.functions import Coalesce
+
+from hitas.models.housing_company import RegulationStatus
+from hitas.models.owner import Owner, OwnerT
+from hitas.models.ownership import Ownership
+from hitas.utils import SQSum
+
+
+def obfuscate_owners_without_regulated_apartments() -> list[OwnerT]:
```
ranta, line 11: Instead of doing so much data processing in python, wouldn't it be easier to let the db handle it?
(Increases db queries by 1, but less time is spent running python code)

```py
owners_qs = Owner.objects.annotate(...).filter(...)

obfuscated_owners: list[OwnerT] = list(owners_qs.values("name", "identifier", "email"))

owners_qs.update(
    name=None,
    identifier=None,
    valid_identifier=False,
    email=None,
    bypass_conditions_of_sale=True,
)
return obfuscated_owners
```

ranta (approved):
🐗

backend/hitas/services/owner.py
matti-lamppu: Yes, this is much cleaner, and if I add an if-check before the queryset.update, it's the same amount of queries since the `obfuscated_owners` is evaluated once, and then the update is depending on if there are obfuscated owners.

```python
if obfuscated_owners:
    owners.update(...)
```

PR #274, opened 2023-04-04
Fix apartment sale sorting when multiple sales are made in the same day

matti-lamppu:
# Hitas Pull Request

# Description

Fixes apartment's latest sale when there are multiple sales on the same day so that the latest sale is from the largest id. Same applies to the first sale, it is from the smallest id if there are multiple on the same day.

> Some of the sortings were the wrong way round, in which case they were corrected.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] Create multiple sales for an apartment on the same day. The most recent one should still be where the latest sale purchase price ("viimeisin kauppakirjahinta") is taken from. Same for the first sale, if there are multiple on the same day, the first one should be the first one made.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-524

### DIFF
--- a/backend/hitas/calculations/max_prices/max_price.py
+++ b/backend/hitas/calculations/max_prices/max_price.py
@@ -320,7 +320,7 @@ def fetch_apartment(
             _first_sale_purchase_price=Subquery(
                 queryset=(
                     ApartmentSale.objects.filter(apartment_id=OuterRef("id"))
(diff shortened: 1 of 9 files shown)


ranta (approved):
🐳

PR #275, opened 2023-04-04
Use empty property manager info when missing

matti-lamppu:
# Hitas Pull Request

# Description

Fixes max price calculation when property manager is missing

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] Try to create a max price calculation in an apartment in a housing company that has no property manager. It should work now.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-523

### DIFF
--- a/backend/hitas/calculations/max_prices/max_price.py
+++ b/backend/hitas/calculations/max_prices/max_price.py
@@ -29,7 +29,7 @@ from hitas.models import (
 )
 from hitas.models._base import HitasModelDecimalField
 from hitas.models.apartment import ApartmentWithAnnotationsMaxPrice
-from hitas.utils import monthify
+from hitas.utils import monthify, safe_attrgetter
 
 
 def create_max_price_calculation(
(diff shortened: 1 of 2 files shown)


backend/hitas/calculations/max_prices/max_price.py
```diff
@@ -216,8 +216,16 @@ def calculate_max_price(
...
+                "name": safe_attrgetter(
+                    apartment.housing_company,
+                    "property_manager.name",
+                    default="",
+                ),
+                "street_address": safe_attrgetter(
+                    apartment.housing_company,
+                    "property_manager.street_address",
+                    default="",
+                ),
```
ranta, line 228: I wonder it would be clearer to have `property_manager` be `None` if it's not set instead of returning a blank string 🤔

ranta (approved):
🕊️

backend/hitas/calculations/max_prices/max_price.py
matti-lamppu: It's simpler this way since code using the property manager info does not need to handle Nones explicitly, e.g. [confirmed_maximum_price.jinja:150](https://github.com/City-of-Helsinki/hitas/blob/8b78560ce01d5c1fd9e06027026977089cde7b8f/backend/hitas/templates/confirmed_maximum_price.jinja#L150).

PR #277, opened 2023-04-18
Add user info endpoint

matti-lamppu:
# Hitas Pull Request

# Description

Allows user information to be fetched for users logged in using Tunnistamo.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [x] Login using Tunnistamo and use the provided endpoint for user information.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-532

### DIFF
--- a/backend/config/urls.py
+++ b/backend/config/urls.py
@@ -3,6 +3,7 @@ from django.contrib import admin
 from django.urls import include, path
 
 from hitas.views.health import HealthCheckView
+from users.views import UserInfoView
 
 urlpatterns = [
     path("admin/", admin.site.urls),
@@ -10,6 +11,7 @@ urlpatterns = [
     path("healthz", HealthCheckView.as_view(), name="health_check"),
     path("pysocial/", include("social_django.urls", namespace="social")),
     path("helauth/", include("helusers.urls")),
+    path("helauth/userinfo/", UserInfoView.as_view(), name="userinfo"),
 ]
 
 if settings.DEBUG_TOOLBAR:
--- a/backend/users/views.py
+++ b/backend/users/views.py
@@ -1,0 +1,29 @@
+import json
+from base64 import urlsafe_b64decode
+
+from rest_framework import status
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from social_django.models import UserSocialAuth
+
+
+class UserInfoView(APIView):
+    def get(self, request: Request, *args, **kwargs) -> Response:
+        try:
+            user_info = UserSocialAuth.objects.get(user=request.user)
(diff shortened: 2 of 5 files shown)


backend/config/urls.py
```diff
@@ -3,13 +3,15 @@
...
 from hitas.views.health import HealthCheckView
+from users.views import UserInfoView
 
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("api/v1/", include("hitas.urls")),
     path("healthz", HealthCheckView.as_view(), name="health_check"),
     path("pysocial/", include("social_django.urls", namespace="social")),
     path("helauth/", include("helusers.urls")),
+    path("helauth/userinfo/", UserInfoView.as_view(), name="userinfo"),
```
ranta, line 14: Even though it is related to helauth, as an API endpoint, this should be under `api/v1/` 🙂

backend/users/views.py
```diff
@@ -0,0 +1,29 @@
...
+        raw_token = user_info.extra_data["id_token"].split(".")[1]
+        padded_token = raw_token + "=" * divmod(len(raw_token), 4)[1]
+        token_payload = json.loads(urlsafe_b64decode(padded_token))
+
+        data = {
+            "first_name": token_payload.get("given_name"),
+            "last_name": token_payload.get("family_name"),
+            "email": token_payload.get("email"),
+        }
+        return Response(data=data, status=status.HTTP_200_OK)
```
ranta, line 29: I would add a fallbacks here.
- If the `UserSocialAuth` object does not have a first or last name, we can get them from our own `User` model.
- Additionally if we want to keep using Django users when testing, we should fallback to using them if `UserSocialAuth.objects.get(user=request.user)` is not found 😉

backend/config/urls.py
matti-lamppu: I disagree. The endpoint should only be used with users authenticated with Helsinki profile, and the endpoint path reflects that. I think of it as an extension to Helsinki profile rather than a Hitas-specific thing. It's also not something that would be versioned with the rest of the endpoints, so I don't see why it would be under the version path from that perspective either.

backend/users/views.py
matti-lamppu: A fallback to `User` model first name, last name, and email make sense, I'll implement that.

Currently, I haven't changed how regular tests create the authenticated user (`UserSocialAuth` model is not created in `hitas.tests.conftest.api_client`). `DEFAULT_AUTHENTICATION_CLASSES` still has both DEV and Helsinki profile tokens on, so tests will continue to work normally until that is changed.

ranta (approved):
🐥

PR #278, opened 2023-04-18
Thirty year regulation changes

matti-lamppu:
# Hitas Pull Request

# Description

Implements discussed changes for 30-year regulation:
- [x] External sales import endpoint uses date instead of a quarter
- [x] 30-year regulation includes more fields
- [x] 30-year regulation results can be fetched without re-checking regulation
- [x] 30-year regulation can only be re-checked if is no previous regulation results, or there are skipped housing companies in the previous results
- [x] Do not allow overriding external sales data
- [x] Add calculation date parameter to external sales data endpoint to validate that the given Excel sheet matched that date.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests

## Tickets

This pull request resolves all or part of the following ticket(s): HT-535, HT-536, HT-538, HT-539, HT-540, HT-541

### DIFF
--- a/backend/hitas/services/thirty_year_regulation.py
+++ b/backend/hitas/services/thirty_year_regulation.py
@@ -14,7 +14,7 @@ from openpyxl import Workbook
 from openpyxl.styles import Alignment, Border, Font, Side
 from openpyxl.worksheet.worksheet import Worksheet
 
-from hitas.exceptions import HitasModelNotFound, get_hitas_object_or_404
+from hitas.exceptions import HitasModelNotFound, ModelConflict, get_hitas_object_or_404
 from hitas.models.apartment_sale import ApartmentSale
 from hitas.models.external_sales_data import ExternalSalesData, SaleData
 from hitas.models.housing_company import (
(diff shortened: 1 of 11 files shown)


backend/openapi.yaml
```diff
@@ -2345,18 +2345,28 @@ paths:
   /api/v1/thirty-year-regulation:
 
     get:
-      description: Check housing companies over 30 years old whether they should be released from hitas regulation or not.
-      operationId: check-regulation
+      description: >-
+        If 'check' parameter is 'true', check housing companies over 30 years old
+        whether they should be released from Hitas regulation or not.
+        Otherwise, try to fetch existing regulation results.
+      operationId: read-or-check-regulation
```
ranta, line 2352: So "check" is actually not checking _(tarkastaa)_, but running the regulation? Maybe some rewording could be in order..

Also it would be better to split running the regulation in the same endpoint from fetching the results. a GET endpoint is generally never used to modify data. I would the "run regulation" apart to a GET and POST endpoint

matti-lamppu: Yes, it would make more sense to split those into two endpoints, I'll implement that.

backend/hitas/services/thirty_year_regulation.py
```diff
@@ -231,6 +233,22 @@ def perform_thirty_year_regulation(calculation_date: datetime.date) -> Regulatio
     return results
 
 
+def check_existing_regulation_data(calculation_month: datetime.date) -> None:
+    """
+    Check if there is already regulation results for this hitas quarter.
+    Allow re-check only if there are skipped housing companies.
```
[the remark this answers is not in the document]
matti-lamppu: The "skipped" list is filled only when the housing company should be checked but couldn't be due to missing prices for its area. I can return early if we get any skipped results, and that would make the logic easier in a few places.

PR #285, opened 2023-05-02
Authentication frontend

annisassi:
# Hitas Pull Request

# Description

Authentication and authorization added to the frontend
    - Sign in and log out -buttons are added to the header
    - Clicking sign in directs the user to the sign in dialog, and after successful sign in, redirects back to where the user left
    - User name or email is shown for authenticated users
    - Protected content is shown only for the authorized users
    - Unauthorized, loading or logout  -views are shown for the unauthorized users

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**
  - [x] Tooltips have been added in the frontend for all new fields
  - [x] Test instructions have been written for the customer in the appropriate ticket in Jira

## Test plan

  - [x] Sign in from the header icon in the frontend, and verify that a sign-in dialog opens and after successful sign in, you are directed back to the page you left
  - [x] Verify that the content you see in the frontend reflects the user rights of your account and the authentication status
  - [x] Verify that your user name or email is visible next to the user icon in the header when signed in
  - [x] Log out from the user dropdown in the header, and verify that you see the successful log out -message, and that you cannot access the protected content anymore. You should see the instructions to sign in instead of the protected content when logged out.
  - [x] Sign in using the hyperlink in the content area of the frontend and verify that it works similar to the sign in icon button in the header

## Tickets

This pull request resolves all or part of the following ticket(s): HT-486, HT-487, HT-529,  HT-530


backend/config/settings.py
```diff
@@ -218,6 +219,8 @@ def relativedelta_months(value: int) -> relativedelta:
 SOCIAL_AUTH_TUNNISTAMO_SECRET = env("SOCIAL_AUTH_TUNNISTAMO_SECRET")
 SOCIAL_AUTH_TUNNISTAMO_SCOPE = ["ad_groups"]
 SOCIAL_AUTH_TUNNISTAMO_AUTH_EXTRA_ARGUMENTS = {"ui_locales": "fi"}
+SOCIAL_AUTH_TUNNISTAMO_ALLOWED_REDIRECT_HOSTS = ["localhost:3000"]
```
matti-lamppu, line 222: This should be set from env, but I can make this fix.

frontend/.env.development
```diff
@@ -0,0 +1,10 @@
+# Docker:
+# REACT_APP_API_DOMAIN=http://localhost:8080 
+# Local:
+REACT_APP_DOMAIN=http://localhost:3000 
+REACT_APP_API_DOMAIN=http://localhost:8000
+REACT_APP_API_BASEURL=$REACT_APP_API_DOMAIN/api/v1
+REACT_APP_AUTH_API_BASEURL=$REACT_APP_API_DOMAIN/helauth
+REACT_APP_AUTH_LOGIN_BASEURL=$REACT_APP_API_DOMAIN/helauth/login?next=
+REACT_APP_AUTH_LOGOUT_BASEURL=$REACT_APP_API_DOMAIN/helauth/logout?next=
+
```
matti-lamppu, line 10: Extra empty line which makes CI fail

tunnistamo
matti-lamppu, line 1: Why is Tunnistamo submodule deleted? I don't think it should be?

annisassi: Tunnistamo folder was added in .gitignore in the master updates. Should that be removed too? I had to delete it from git to avoid conflicts.

frontend/.env.development
annisassi: Removed the extra lines

tunnistamo
matti-lamppu: The folder was added to `.gitignore` so that changes inside it wouldn't be shown in git diffs. The folder should still be included, because git submodules do not work correctly, if the submodule folder is not committed to the repository. You can verify this by cloning hitas and running `git submodule init` and `git pull --recurse-submodules`, which will correctly init the tunnistamo submodule on the master branch but not on this branch.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "tunnistamo"]
 	path = tunnistamo
-	url = https://github.com/City-of-Helsinki/tunnistamo
+	url = ./tunnistamo/
```
matti-lamppu, line 3: I don't think this works. You should add the submodule back with `git submodule add https://github.com/City-of-Helsinki/tunnistamo`

.gitignore
```diff
@@ -130,7 +130,7 @@ dmypy.json
 
 .idea/
 *.iml
-tunnistamo/
+tunnistamo/docker-compose.env.yaml
```
matti-lamppu, line 133: The whole tunnistamo folder should be excluded.

backend/config/settings.py
matti-lamppu: Done

matti-lamppu:
Login and logout work as expected. Also tested with remote Tunnistamo and it also works great 👍

matti-lamppu:
@vincit-matu or @ranta should review the frontend code

ranta (changes requested):
If the .env variable `REACT_APP_AUTH_TOKEN=...` is set, clicking on 'Kirjaudu sisään' sends multiple requests, but the app is forever requesting the user to authenticate.

annisassi:
> If the .env variable `REACT_APP_AUTH_TOKEN=...` is set, clicking on 'Kirjaudu sisään' sends multiple requests, but the app is forever requesting the user to authenticate.

I think this is likely to be related to backend settings as the token way of authentication has not been changed in the frontend.  The backend now directs to tunnistamo login also with token and sets the cookie, which is then not used in the queries as the token is sent instead. Thus the request from userinfo endpoint returns 404 not found. Frontend considers the user logged in only if userinfo data is returned from the backend.

tunnistamo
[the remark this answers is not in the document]
annisassi: Restored tunnistamo

frontend/.env.production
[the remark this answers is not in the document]
annisassi: Combined the separate env files to one .env file.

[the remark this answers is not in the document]
annisassi: Now the token authentication should also work...

frontend/src/app/services.ts
```diff
@@ -91,14 +91,34 @@ export const hitasApi = createApi({
...
+
+const userApi = authApi.injectEndpoints({
+    endpoints: (builder) => ({
+        getUserInfo: builder.query<IUserInfoResponse, void>({
+            query: () => ({
+                url: "userinfo",
+            }),
+        }),
+    }),
+});
```
ranta, line 120: Right now the userInfo is not cached at all, which means that on every page change the frontend queries the API to get the userInfo. We can cache the request's results with tags like so:

```ts
export const authApi = createApi({
    ...
    tagTypes: ["UserInfo"],
    endpoints: (builder) => ({
        getUserInfo: builder.query<IUserInfoResponse, void>({
            query: () => ({
                url: "userinfo",
            }),
            providesTags: ["UserInfo"],
        }),
    }),
});
```

---

After we have cached the data and as the userInfo pretty much never changes, we can extend the default cache duration with `keepUnusedDataFor` so we should need to fetch the userInfo only once per authentication
https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#reducing-subscription-time-with-keepunuseddatafor

---

After we have the cache enabled, you should then invalidate the cache when the user logs out, so that the userInfo is not shown  after user is already logged out.
https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#resetapistate

ranta (changes requested):
Getting really close already, authentication seems to work great with Tunnistamo and the `REACT_APP_AUTH_TOKEN` env variable. Still a few fixes needed before we can merge this :)

I highly recommend rebasing PRs to `master` any time you notice there have been changes made to `master`

.gitignore
```diff
@@ -103,6 +103,7 @@ celerybeat.pid
 
 # Environments
 .env
+!frontend/.env
```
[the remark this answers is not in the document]
annisassi: I used separate development env file originally for the token so that it is not included to the production build as it should not go there at all. So the intention was to have one env file for every build and one local env file for token in only development builds. The current simplification does not do this anymore. I think I should move the current .env.local back to .env.development.local file. I see there is .env and .env.template in the backend, but in that pattern the token would easily end up to the production build.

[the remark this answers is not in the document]
annisassi: It's changed to use only .env and .env.template files

frontend/src/app/services.ts
annisassi: I tested this with the tags and keepUnusedDataFor but these do not fix the problem. It should cache the data by default without tags and keep the cache on App component, which is not unmounted. I don't know why it still seems to call the endpoint on some route changes instead of using the cache.

ranta: 👍 Not a major issue, we can resolve that one later if its needed

ranta (approved):

PR #288, opened 2023-05-04
Endpoint for regulation postal codes

matti-lamppu:
# Hitas Pull Request

# Description

Adds the ability to fetch postal codes with their price per square meter and cost area, so replacement postal codes can be added if some housing companies cannot be regulated due to a lack of sales data.

Also fixes a bug in thirty-year regulation where if regulation is made in the last month of the Hitas-quarter, the range of sales dates would be off.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] Create external sales data and try the new endpoint.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-542

### DIFF
--- a/backend/hitas/tests/apis/test_api_thirty_year_requlation_postal_codes.py
+++ b/backend/hitas/tests/apis/test_api_thirty_year_requlation_postal_codes.py
@@ -1,0 +1,134 @@
+import datetime
+
+import pytest
+from dateutil.relativedelta import relativedelta
+from django.urls import reverse
+from rest_framework import status
+
+from hitas.models import Apartment, ExternalSalesData
+from hitas.models.external_sales_data import CostAreaData, QuarterData
+from hitas.models.housing_company import HitasType, RegulationStatus
+from hitas.tests.apis.helpers import HitasAPIClient, count_queries
+from hitas.tests.factories import ApartmentFactory, ApartmentSaleFactory
+from hitas.utils import to_quarter
+
+
+@pytest.mark.django_db
+def test__api__regulation_postal_codes(api_client: HitasAPIClient, freezer):
+    day = datetime.datetime(2023, 2, 1)
+    freezer.move_to(day)
+
+    this_month = day.date()
+    previous_year_last_month = this_month - relativedelta(months=2)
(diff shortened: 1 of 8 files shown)


backend/hitas/tests/apis/test_api_thirty_year_requlation_postal_codes.py
```diff
@@ -0,0 +1,134 @@
...
+    assert response.json() == [
+        {
+            "postal_code": "00001",
+            "price_by_area": 49000.0,
+            "cost_area": 1,
+        },
+        {
+            "postal_code": "00002",
+            "price_by_area": 15000.0,
+            "cost_area": None,
```
ranta, line 76: In what cases would the `cost_area` ever be `None`? It is a required field on the model after all. 
If it's a Postal Code that's not in our system, it can be ignored here for not being relevant to Hitas (Unless I remember wrong?) 🤔

matti-lamppu: I meant to ask about this: Do we have PostalCode entities for all possible postal codes in the Helsinki area, and not just the ones in the Hitas system? External sales data can contain sales in postal codes that do not have any Hitas apartments, but the user can still select these if there is enough sales data for that region. If we only have entities for Hitas housing companies, we could end up with a None for cost area.

matti-lamppu: https://github.com/City-of-Helsinki/hitas/pull/288/files#diff-e56b63a597e48faaf0e9b2c178934604c9a3e0e5b75ab71258b4843947bf7e16R855

ranta: > External sales data can contain sales in postal codes that do not have any Hitas apartments, but the user can still select these

But don't have a restriction that if there is a Hitas postal code with no sales, the user must pick a postal code that has the same `cost_area`?

ranta (approved):
🐏

PR #300, opened 2023-05-10
Thirty year comparison front

vincit-matu:
# Hitas Pull Request

# Description

Adds the functionality and styles for the thirty year comparison, along with testing capabilities.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

Go to "Toiminnot" page, and the "Vapautumisen tarkistus" tab within it. If you select "Testi" from the year selector, the other dropdown will be populated with the different test cases (as well as a case for a successful comparison). Changing the test case will also reset the page, so you can test them all in sequence if you want.
Since we don't have the price ceiling calculation done you can't test the comparison "for real" yet, but the "Onnistunut vertailu" will simulate the same kind of API-response as you'd get from a successful one.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-484, it also has a lot of the frontend done for the other tickets related to the thirty year comparison.


ranta (commented):
The “Aloita Vertailu” button does not get enabled after an excel file is successfully uploaded. Only after changing the date to something else, then coming back is it enabled.

---

Improvement suggestion: Remove the “Tallenna keskineliöhinnat” button, instead when a file is selected do the API request.

---

Something with the dates seems a bit messed up?
- When doing the calculation for e.g. 2022Q3, The external sales data that backed uses for the last 4 quarters comes from the 2022Q2 file, so if you haven't uploaded that already the calculation fails due to missing external sales data.
- However if you upload the 2022Q3 sales data the UI allows the "Aloita Vertailu" to be clicked

---

Seems to be missing loading indicator when changing the date

---

When the 30 year regulation is created for a month, the data returned from the API is not shown in the frontend, instead it allows the user to run the regulation again which returns a 409 error from the API

---

When the "Aloita Vertailu" is clicked, and it returns a successful response (200), an error message pops up that the results cant be found (The modal is also full height for some reason)
image

ranta (approved):
Comments will be addressed in later PRs, we can merge this already :)

PR #311, opened 2023-05-15
Conditions of sale should be created after ownerships in initial.json

matti-lamppu:
# Hitas Pull Request

# Description

Initial.json is broken so here is a fix.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

Try to load backend with the changed initial.json and see if it works

## Tickets

This pull request resolves all or part of the following ticket(s): -

### DIFF
--- a/backend/initial.json
+++ b/backend/initial.json
@@ -2680,30 +2680,6 @@
     }
   },
   {
-    "model": "hitas.conditionofsale",
-    "pk": 1,
-    "fields": {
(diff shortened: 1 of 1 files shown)


matti-lamppu:
I think make dump tries to format the dump alphabetically, which is not correct, since it breaks dependencies like this. If you just use the `manage.py dumpdata` then it should work as expected.

PR #328, opened 2023-05-25
Switch to django-post-fetch-hook for obfuscation hooks

matti-lamppu:
# Hitas Pull Request

# Description

Switched to [django-post-fetch-hook](https://github.com/MrThearMan/django-post-fetch-hook) for post save obfuscation hooks (library based on the implementation previously in Hitas). More robust and 100% tested.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests

## Tickets

This pull request resolves all or part of the following ticket(s): HT-599

### DIFF
--- a/backend/hitas/models/owner.py
+++ b/backend/hitas/models/owner.py
@@ -38,40 +38,46 @@ class Owner(ExternalSafeDeleteHitasModel):
         super().save(*args, **kwargs)
 
     @classmethod
-    def post_fetch_hook(cls, results, fields):
-        for i, owner in enumerate(results):
-            # qs.values(...)
-            if isinstance(owner, dict):
-                # If non_disclosure is not given, we assume the owner needs to be obfuscated
-                if "non_disclosure" not in owner or owner["non_disclosure"] is True:
-                    for field in cls.obfuscation_rules:
-                        if field in owner:
-                            owner[field] = None
+    def post_fetch_hook(cls, model: "Owner") -> "Owner":
+        obfuscate(model)
+        return model
 
-            # qs.values_list(...)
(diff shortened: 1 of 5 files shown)


backend/hitas/models/owner.py
```diff
@@ -38,40 +38,39 @@ def save(self, *args, **kwargs) -> None:
...
+                if field in values:
+                    values[field] = value
+        return values
+
+    @classmethod
+    def post_fetch_values_list_hook(cls, values: tuple[Any, ...], fields: tuple[str, ...]) -> tuple[Any, ...]:
+        # Fields that were accessed
+        index_nd = index_of(fields, "non_disclosure")
+
+        # If non_disclosure is not given, we assume the owner needs to be obfuscated
```
ranta, line 59: I'm not a fan of *assuming* that everything needs to be obfuscated if `non_disclosure` field is missing, instead i'd rather:
- Raise an exception or 
- Add the `non_disclosure` field to the values, and later pop it out

matti-lamppu: I do the add-and-pop method already here: 

https://github.com/City-of-Helsinki/hitas/blob/39c7f5add9205fe306b4281e01ecb1546f3861f7/backend/hitas/services/owner.py#L39 

Maybe making it mandatory to include by raising a RuntimeExcepition could be a better way?

ranta: Hmm yes, then we would need to do less 'hacky' stuff 😄

matti-lamppu: Done.

PR #339, opened 2023-05-30
Convert apartment state into dynamic propery

matti-lamppu:
# Hitas Pull Request

# Description

Apartment state is now determined based on if the apartment has any sales or not. There are only two states now, "free" and "sold". State is removed from apartment and cannot be modified directly.

Frontend needs to be changed to not display apartment state in apartment edit page.

## Pull request checklist

- **Testing**
- **Database**
  - [x] Database migrations will work in the DEV & TEST environments
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] Check that state is returned on the endpoint, but that it cannot be changed anymore

## Tickets

This pull request resolves all or part of the following ticket(s): HT-192


backend/hitas/models/apartment.py
```diff
@@ -30,12 +30,10 @@
 
 class ApartmentState(Enum):
     FREE = "free"
-    RESERVED = "reserved"
     SOLD = "sold"
 
     class Labels:
         FREE = _("Free")
-        RESERVED = _("Reserved")
         SOLD = _("Sold")
```
ranta, line 37: Since we are changing this from a '3 value enum' to a enum with only 2 values, it doesn't make sense to keep it as an enum anymore. I would do a more major and change `state` to a `is_sold` (boolean), as even the `state` name does not fit that well anymore

ranta (changes requested):

backend/hitas/models/apartment.py
matti-lamppu: I initially had this a boolean but decided to try to minimize required changes in the frontend, and so I kept this as an enum. We can go with the boolean route though.

ranta: If you want, I can do the frontend part to this same PR when the backend is using booleans, just ping me :)

matti-lamppu: Backend changes to `is_sold` now in this branch, go ahead with the frontend!

PR #341, opened 2023-05-31
Add sales report endpoint

matti-lamppu:
# Hitas Pull Request

# Description

Adds endpoint for creating and downloading sales report Excel file.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests
- [ ] Uncomment the lines below from the top of `backend/openapi.yaml` and open the Swagger interface at `localhost:8090`. Add `http://localhost:8090` to `CORS_ALLOWED_ORIGINS` in `backend/.env`. Add `authentication_classes = []` & `permission_classes = []` to `hitas.views.reports.SalesReportView`. Use the new report endpoint from the Swagger interface while the backend is running at `localhost:8000` and observe the downloaded Excel file.

```yaml
#servers:
#  - url: http://localhost:8000
#    description: Local development server
```

## Tickets

This pull request resolves all or part of the following ticket(s): HT-606

### DIFF
--- a/backend/hitas/services/reports.py
+++ b/backend/hitas/services/reports.py
@@ -1,0 +1,212 @@
+import datetime
+import string
+from decimal import Decimal
+from functools import cache
+from statistics import mean
+from typing import Any, Callable, Iterable, NamedTuple, TypeVar
+
+from openpyxl.styles import Alignment, Border, Side
+from openpyxl.workbook import Workbook
+from openpyxl.worksheet.worksheet import Worksheet
+from rest_framework.exceptions import ValidationError
+from rest_framework.settings import api_settings
+
+from hitas.models import ApartmentSale
+from hitas.utils import format_sheet, resize_columns
+
+
+class SalesReportColumns(NamedTuple):
+    cost_area: int | str
+    postal_code: str
+    apartment_address: str
+    notification_date: datetime.date | str
+    purchase_date: datetime.date | str
+    purchase_price_per_square_meter: Decimal | str
+    total_price_per_square_meter: Decimal | str
+
+
+T = TypeVar("T")
+
+
+class SalesReportSummaryDefinition(NamedTuple):
+    func: Callable[[Iterable[T]], T]
+    title: str = ""
+    subtitle: str = ""
+
+
+def build_sales_report_excel(sales: list[ApartmentSale]) -> Workbook:
+    workbook = Workbook()
+    worksheet: Worksheet = workbook.active
+
+    column_headers = SalesReportColumns(
+        cost_area="Kalleusalue",
+        postal_code="Postinumero",
+        apartment_address="Osoite",
+        notification_date="Ilmoituspäivä",
+        purchase_date="Kauppapäivä",
+        purchase_price_per_square_meter="Kaupan neliöhinta",
+        total_price_per_square_meter="Velaton neliöhinta",
+    )
+    worksheet.append(column_headers)
+
+    for sale in sales:
+        if not sale.apartment.surface_area:
+            raise ValidationError(
+                detail={
+                    api_settings.NON_FIELD_ERRORS_KEY: (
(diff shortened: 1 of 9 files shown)


backend/hitas/views/reports.py
```diff
@@ -0,0 +1,38 @@
...
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        if attrs["start_date"] > attrs["end_date"]:
+            raise serializers.ValidationError("Start date must be before end date")
+        return attrs
+
+
+class SalesReportView(ViewSet):
+    renderer_classes = [HitasJSONRenderer, ExcelRenderer]
+
+    def create(self, request: Request, *args, **kwargs) -> HttpResponse:
```
ranta, line 28: Don't all of our other PDF endpoints use GET instead of POST? 🤔 I'd like to keep things consistent if possible :)

matti-lamppu: This is an Excel endpoint, but `download-regulation-results` Excel endpoint also uses GET, so I can convert this to get for consistency. 

Although this endpoint has all of its parameters required when the other endpoint has only optional parameters which suit query parameters better..

matti-lamppu: Done

matti-lamppu:
Still needs approval from the customer before merging, unless we wanna do an update later

backend/hitas/views/reports.py
```diff
@@ -0,0 +1,38 @@
...
+from hitas.views.utils.excel import ExcelRenderer, get_excel_response
+
+
+class SalesReportSerializer(serializers.Serializer):
+    start_date = serializers.DateField()
+    end_date = serializers.DateField()
+
+    def validate(self, attrs: dict[str, Any]) -> dict[str, Any]:
+        if attrs["start_date"] > attrs["end_date"]:
+            raise serializers.ValidationError("Start date must be before end date")
```
ranta, line 21: 🤔

ranta: We can get this merged for now, and fix it in some later PR to avoid rebasing all of them 😄

backend/hitas/services/reports.py
```diff
@@ -0,0 +1,227 @@
...
+        worksheet.append(
+            SalesReportColumns(
+                cost_area=sale.apartment.postal_code.cost_area,
+                postal_code=sale.apartment.postal_code.value,
+                apartment_address=sale.apartment.address,
+                notification_date=sale.notification_date,
+                purchase_date=sale.purchase_date,
+                purchase_price_per_square_meter=sale.purchase_price / sale.apartment.surface_area,
+                total_price_per_square_meter=sale.total_price / sale.apartment.surface_area,
+                exclude_from_statistics=sale.exclude_from_statistics,
```
ranta, line 74: I would change the "Ei Tilastoihin" value from "TRUE/FALSE" to something more descriptive. e.g. Only showing TRUE values

ranta (approved):
I have some formatting issues with dates, but that most likely just a mac Sheets app problem 😅

backend/hitas/views/reports.py
matti-lamppu: Isn't this correct? We are giving a range to search sales from, so the start of the range needs to be before the end of the range (or the same, if you just want one day).

backend/hitas/services/reports.py
matti-lamppu: I can put X / " " there also.

matti-lamppu: Actually, we changed this already so that excluded sales are not in this report, based on the customer's feedback, so this column should be removed.

matti-lamppu: Field removed

backend/hitas/views/reports.py
ranta: Yes nvm, brain fart 😅

PR #356, opened 2023-06-08
front: Owners list in codes tab & fixes to modify owner info modal

annisassi:
# Hitas Pull Request

# Description

- Added owners search to the owners tab in codes. Clicking the row in the search results opens the owner info modification modal.
- Fixed validation issue, completed validation rules to align with the backend validation rules and set the initial focus for the modify owner info modal.
- Ensured the owners list updates after modifying the owner info.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

Open the owner search from the owners tab in codes. Ensure you can search owners with part of the name and/or identifier. Click a row in the search results list and check that the initial focus and validations work as expected in the owner modification modal. Modify and save the owner information and check that the data in the search results list updated correspondingly.

## Tickets

This pull request resolves all or part of the following ticket(s): HT-588

### DIFF
--- a/frontend/src/common/components/ModifyOwnerModal.tsx
+++ b/frontend/src/common/components/ModifyOwnerModal.tsx
@@ -55,8 +55,8 @@ const OwnerMutateForm = ({owner, closeModalAction}: IOwnerMutateForm) => {
     const identifierValue = ownerFormObject.watch("identifier");
 
     // helper booleans for special saving controls with invalid data
-    const hasFormChanged = Object.entries(owner).toString() !== Object.entries(ownerFormObject.getValues()).toString();
-    const isMalformedIdentifier = ownerFormObject.formState.errors.identifier?.type === "custom";
+    const hasFormChanged = ownerFormObject.formState.isDirty;
+    const isMalformedIdentifier = !!ownerFormObject.formState.errors.identifier;
     const isIdentifierEmpty = identifierValue === "";
     const isBackendErrorInIdentifier = ownerFormObject.formState.errors.identifier?.type === "backend";
     const isOtherErrorsThanIdentifier = Object.keys(ownerFormObject.formState.errors).some(
@@ -75,9 +75,12 @@ const OwnerMutateForm = ({owner, closeModalAction}: IOwnerMutateForm) => {
         !hasFormChanged;
 
     useEffect(() => {
-        // validate the initial form values and set the initial identifier validity
+        // validate the initial form values
         ownerFormObject.trigger().then(() => {
+            // set the initial identifier validity
             setIsInitialIdentifierValid(!ownerFormObject.formState.errors.identifier);
+            // set initial focus
+            setTimeout(() => ownerFormObject.setFocus("name"), 5);
         });
         // eslint-disable-next-line
     }, []);
--- a/frontend/src/features/codes/OwnersList.tsx
+++ b/frontend/src/features/codes/OwnersList.tsx
@@ -1,0 +1,113 @@
+import {IconCrossCircle, IconSearch, TextInput} from "hds-react";
+import {useCallback, useRef, useState} from "react";
+import {useGetOwnersQuery} from "../../app/services";
+import {ModifyOwnerModal, QueryStateHandler} from "../../common/components";
+import {IFilterOwnersQuery, IOwner} from "../../common/schemas";
+
+const OwnerResultList: React.FC<{params: IFilterOwnersQuery}> = ({params}) => {
+    const MIN_LENGTH = 2; // Minimum length before applying filter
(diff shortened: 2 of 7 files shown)


frontend/src/styles/components/_Codes.sass
```diff
@@ -101,3 +100,8 @@
     @include flexbox()
     @include justify-content(center)
     padding-top: $spacing-layout-m
+
+  .list-footer
+      margin-left: auto
+      margin-right: auto
+      margin-top: 1em
```
ranta, line 107: To stop GitHub from warning about this, you should enable 'Ensure every saved file ends with a line break' setting in your IDE.

frontend/src/common/components/ModifyOwnerModal.tsx
```diff
@@ -52,12 +52,12 @@ const OwnerMutateForm = ({owner, closeModalAction}: IOwnerMutateForm) => {
         mode: "all",
         resolver: resolver,
     });
-    const identifierValue = ownerFormObject.watch("identifier");
+    const ownerFormObjectValues = ownerFormObject.watch();
```
ranta, line 55: If you only need one value from the formObject, you can do `ownerFormObject.watch("identifier");` 😉

frontend/src/features/codes/OwnersList.tsx
```diff
@@ -0,0 +1,113 @@
...
+            />
+        </QueryStateHandler>
+    );
+};
+
+export default function OwnersList() {
+    // search strings
+    const [filterParams, setFilterParams] = useState<IFilterOwnersQuery>({name: "", identifier: ""});
+
+    // focus the field when clicking its cross circle icon
```
ranta, line 69: Ooh, this is a nice one 😄 
You should consider adding this to `FilterTextInputField` also!

frontend/src/features/codes/OwnersList.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <TextInput
+                    id="filter__name"
+                    ref={nameSearchRef}
+                    label="Nimi"
+                    value={filterParams.name}
+                    onChange={(e) => setFilterParams((prev) => ({...prev, name: e.target.value}))}
+                    onButtonClick={() => clearAndFocus("name")}
+                    autoFocus
+                    buttonIcon={filterParams.name ? <IconCrossCircle /> : <IconSearch />}
+                />
```
ranta, line 99: Why not use `FilterTextInputField` here? 🤔

frontend/src/features/codes/OwnersList.tsx
```diff
@@ -0,0 +1,113 @@
...
+            isLoading={isLoading}
+        >
+            <div className="list-headers">
+                <div>Nimi</div>
+                <div>Henkilö- tai Y-tunnus</div>
+            </div>
+            <ul className="results-list">
+                {data?.contents.map((owner: IOwner) => (
+                    <div
+                        key={owner.identifier}
```
ranta, line 41: `owner.identifier` is a bad value for the `key`, as it may be duplicate or `null`. If you have owners without an identifier this listing will break.

frontend/src/styles/components/_Codes.sass
annisassi: Done :)

frontend/src/common/components/ModifyOwnerModal.tsx
annisassi: Watching all the fields was actually needed for the hasFormChanged to update. I changed it to use hook form's isDirty boolean, so I could change it back to watch only one field.

frontend/src/features/codes/OwnersList.tsx
annisassi: To add this feature in FilterTextInputField, it should be changed to a controlled component, that would have an impact to the other components using it (e.g. queries that use min length). So I think this change is a bigger one and should be done in another PR.

frontend/src/features/codes/OwnersList.tsx
annisassi: The comment above applies to this too.

frontend/src/features/codes/OwnersList.tsx
annisassi: Good point. I changed the key to use index instead.

ranta: Index is certainly better than `owner.identifier`, but `owner.id` would be even better 😉 

Using Index as the key is react's way of doing things if key is not provided at all, but it has some issues as well https://react.dev/learn/rendering-lists#why-does-react-need-keys

annisassi: Changed it once more...

PR #373, opened 2023-06-16
Add email sending endpoints

matti-lamppu:
# Hitas Pull Request

# Description

Adds endpoints for sending pdfs via email to given recipients. Recipient emails can be fetched from any source and compiled in the frontend.

> Note that email settings have not been configured yet, so the endpoints do not work. Automated tests have been created that mock out the sending part. Further testing is required after HT-146 is resolved.

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**

## Test plan

- [x] Automated tests

## Tickets

This pull request resolves all or part of the following ticket(s): HT-627, HT-629, HT-633, HT-631


backend/hitas/views/email.py
```diff
@@ -0,0 +1,80 @@
...
+    send_confirmed_max_price_calculation_email,
+    send_regulation_letter_email,
+    send_unconfirmed_max_price_calculation_email,
+)
+from hitas.views.utils import UUIDField
+
+
+class ConfirmedMaxPriceCalculationEmailSerializer(Serializer):
+    calculation_id = UUIDField()
+    template_name = serializers.CharField(max_length=256)
```
ranta, line 19: Should `template_name`s in the Serializers be a `FK`s instead?

Also `EmailTemplateViewSet` could really use a FilterSet :)

backend/hitas/views/email.py
```diff
@@ -0,0 +1,80 @@
...
+    recipients = serializers.ListField(child=serializers.EmailField(), min_length=1)
+
+
+class UnconfirmedMaxPriceCalculationEmailSerializer(Serializer):
+    apartment_id = UUIDField()
+    calculation_date = serializers.DateField()
+    additional_info = serializers.CharField(allow_blank=True)
+    template_name = serializers.CharField(max_length=256)
+    recipients = serializers.ListField(child=serializers.EmailField(), min_length=1)
```
ranta, line 28: Do we need a 'date this was requested' field here too? Assuming we want to track the 'work performance' here also

backend/hitas/views/email.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class RegulationLetterEmailViewSet(ViewSet):
+    def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        serializer = RegulationLetterEmailSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        data = serializer.validated_data
+        send_regulation_letter_email(
+            housing_company_id=data["housing_company_id"],
+            calculation_date=data["calculation_date"],
```
ranta, line 75: I don't think the `calculation_date` does not need to be sent from the frontend, as we can just retrieve it from the 30 year regulation

backend/hitas/services/email.py
```diff
@@ -0,0 +1,243 @@
...
+
+def render_unconfirmed_max_price_calculation_pdf(
+    apartment_id: UUID,
+    calculation_date: datetime.date,
+    additional_info: str,
+    user: User,
+) -> tuple[str, bytes]:
+    calculation_month = monthify(calculation_date)
+
+    apartment: Optional[ApartmentWithAnnotations] = (
```
ranta, line 109: I don't like the amount of duplication **here** vs `ApartmentViewSet` 🙈

backend/hitas/views/email.py
matti-lamppu: I think the name is fine since the flow of sending an email is 

1. Search for the template you want to use for the given type of action, e.g. search for confirmed max price calc templates, and pick one (there shouldn't be that many of them, so a filter set shouldn't be required).
2. Send the email, giving the name of the template you selected.

backend/hitas/urls.py
```diff
@@ -59,6 +59,21 @@
...
+router.register(
+    r"send-unconfirmed-max-price-calculation-email",
+    views.UnconfirmedMaxPriceCalculationEmailViewSet,
+    basename="unconfirmed-max-price-calculation-email",
+)
+router.register(
+    r"send-regulation-letter-email",
+    views.RegulationLetterEmailViewSet,
+    basename="regulation-letter-email",
+)
```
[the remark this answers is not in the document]
matti-lamppu: Will do.

backend/hitas/views/email.py
matti-lamppu: Not sure, will ask.

backend/hitas/views/email.py
matti-lamppu: The calculation date is there to enable sending older regulation letters. Maybe this wont be required, or we only ever want to send from the latest regulation. Will ask.

backend/hitas/services/email.py
matti-lamppu: Yes, I know. 😫 The issue here was just that refactoring and simplifying the queryset and PDF template is a lot of work, since there isn't a good interface in place (it's reusing the apartment endpoint interface, which has a lot of unnecessary data, which is also formatted in a custom way). Maybe it's still worth doing, even with the limited time.

miqu (approved):

PR #376, opened 2023-06-17
Apartment sale for past dates

ranta:
# Hitas Pull Request

# Description

Nyt siinä vaiheessa että kauppasivu käyttää viimeiseksi luotua enimmäishintalaskelmaa mikäli se on kauppapäivänä voimassa, jos se ei ole niin kauppa kykenee hakemaan rajapinnasta rajaneliöhinta-arvion menneisyyden päivämäärälle.
Tämän ohessa tuli tehyä myös kasa pienempiä parannuksia ja korjauksia

## Pull request checklist

- **Testing**
- **Database**
- **Documentation**
  - [x] Tooltips have been added in the frontend for all new fields
  - [x] Test instructions have been written for the customer in the appropriate ticket in Jira

## Tickets

This pull request resolves all or part of the following ticket(s): HT-624


backend/hitas/views/apartment.py
```diff
@@ -922,6 +922,18 @@ def get_detail_queryset(self):
             sapc=self.select_sapc(calculation_month=calculation_month),
         )
 
+    @action(detail=True, methods=["GET"], url_path="retrieve-unconfirmed-prices-for-date")
+    def retrieve_unconfirmed_prices_for_date(self, request, **kwargs):
+        input_data = UnconfirmedPriceSerializer(data=request.query_params)
```
matti-lamppu, line 927: Shouldn't this fetch from confirmed, not unconfirmed max price calculations?

miqu: It might make sense to show on UI also what affects what somehow. And I suspect the same the confirmed price should be searched from the date chosen for sale date and if not available the limitprice from that date used.

ranta: Nope, it is intentionally using unconfirmed prices, as it should be able to use the surface area price ceiling for past sales also. Frontend uses the latest confirmed calculation if it is valid on the sale date, if we need to use an no-most-recent confirmed calculation thats a new feature (also commented this on the jira ticket)

matti-lamppu: Discussed this with the customer, and they said that a new confirmed max price calculation should be made for the given date if one does not already exist since re-sales should always be made using them according to Hitas rules. Therefore we changed the apartment endpoint to fetch the most recent max price calculation before a given date, which also fixes a small issue in the apartment view that would show the most recently confirmed max price, even if there is one which is valid for longer. Frontend changes are still required to make this work.

PR #425, opened 2023-09-19
Non disclosure changes

jforsman:
Updates to to tool version in pre-commit.config

Backend changes for non_disclosure owners:
   HT-660: Do not show non_disclosed owner names in reports, both in pdf and excel.
   HT-671: owners api now return non-obfuscated data even for non_disclosed owners.


annisassi (approved):
HT-671: owners api now return non-obfuscated data even for non_disclosed owners. 
==> Works great
HT-660: Do not show non_disclosed owner names in reports, both in pdf and excel. 
==> This works fine on Janne's computer, I got permission error, so couldn't test it on my computer (propably dev env issue)

